PR #3, opened 2011-07-13
Implement packet listeners

Nightgunner5:
Please note that I haven't tested this because I'm tired and I want to go to bed too

### DIFF
--- a/src/org/bukkitcontrib/ContribNetServerHandler.java
+++ b/src/org/bukkitcontrib/ContribNetServerHandler.java
@@ -23,6 +23,8 @@ import org.bukkitcontrib.inventory.ContribCraftItemStack;
 import org.bukkitcontrib.inventory.ContribCraftingInventory;
 import org.bukkitcontrib.inventory.ContribInventory;
 import org.bukkitcontrib.inventory.CraftingInventory;
+import org.bukkitcontrib.packet.CorePacketType;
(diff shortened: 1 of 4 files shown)


Afforess:
I will test this this evening, to find out exactly how much it impacts performance.

Afforess:
Any chance you can avoid completely destroying the formatting in the existing NSH code?

Nightgunner5:
Stupid Eclipse

Afforess:
Much better. I'm looking at it now.

desmin88:
Yes! +9000 internets.

Afforess:
After review, I am deciding against this.

2 main reasons:

1.) Performance. There is a lot of traffic, and this hurts the biggest servers the hardest. Not good.
2.) This listens to INCOMING packets from the client, not outgoing packets from the server. Most incoming packets already have some sort of event attached to them, I fail to see the need to block them. 

@desmin88 what you need for your project is an outgoing packet listener, so you can block the player specific packets from going to clients from the server. This pull is useless for you.

Nightgunner5:
Sorry. I'm not an expert on packets in Minecraft. Which function would I override to listen for outgoing packets?

Afforess:
NetServerHandler.sendPacket(Packet packet)

Nightgunner5:
I should have asked sooner. It would have saved me a ton of work!

PR #31, opened 2011-07-16
Parses kick message and automatically reconnects to the other server

Raphfrk:
This is pretty much the exact code that I use for patching the client to allow reconnects for ServerPort.

The main change is that both "[Redirect]" and "[Serverport]" can be used as the prefix in the kick message, instead of just "[Serverport]".  That was a request from the Transporter author, so he wouldn't have to hard-code Serverport into his code (which he did in the end, since I was to slow).

ReconnectManager handles the actual parsing of the kick message and also reconnecting.

2 hooks are added to GuiConnectFailed.  The one in the constructor parses the messages and stores the results and the one in drawScreen actually teleports.

This should be compatible with Transporter as well as ServerPort.

### DIFF
--- a/MCP src/GuiConnectFailed.java
+++ b/MCP src/GuiConnectFailed.java
@@ -1,0 +1,58 @@
+package net.minecraft.src;
+
+import net.minecraft.src.GuiButton;
+import net.minecraft.src.GuiMainMenu;
+import net.minecraft.src.GuiScreen;
+import net.minecraft.src.StringTranslate;
+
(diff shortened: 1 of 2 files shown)


Raphfrk:
Btw, when I compile the client, the textures for the portals don't work right.  This is probably just an issue with my build environment, since it happens with and without these changes.

Afforess:
Portal issue is on my end. You didn't do anything to cause it.

PR #32, opened 2011-07-17
Server/Client matching packet lengths

Raphfrk:
This matches the packet lengths for server and client.  

The server uses getNumBytes() via a(), which calls getNumBytes().  However, the client adds 8 to getNumBytes() and returns that as getPacketSize().

This may not be the cleanest fix.  Another option would be to change writePacketData() so that it calls getNumBytes() and leave getPacketSize() return the full packet size.

### DIFF
--- a/MCP src/CustomPacket.java
+++ b/MCP src/CustomPacket.java
@@ -46,7 +46,7 @@ public class CustomPacket extends Packet{
 	 public void writePacketData(DataOutputStream output) throws IOException {
 		  //System.out.println("Writing Packet Data for " + packet.getPacketType());
 		  output.writeInt(packet.getPacketType().getId());
-		  output.writeInt(getPacketSize());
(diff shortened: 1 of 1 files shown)


Raphfrk:
Updated so that it subtracts 8 for the payload length field and getPacketSize() is back to giving the full packet size.

Afforess:
I don't see any changes to getPacketSize()

Raphfrk:
Sorry, the original pull changed line 25 to "return packet.getNumBytes();", so the reduction by 8 happened there.

The updated version has the reduction happening in WritePacketData.  That way getPacketSize() is correct (which since it is a method for Packet probably should be the full length).

Raphfrk:
Ofc, this all depends on what the length field was supposed to mean.

For the server, it gives payload length, but for the client, it gave packet length (so payload + 8).

PR #34, opened 2011-07-17
Update world seed on teleport

Raphfrk:
This adds a "newSeed" variable to the WorldInfo object.  If a Respawn packet is received by the client, it uses the newSeed value as the seed for the new world.

When a player teleport event happens, a packet is sent and it updates the newSeed value.  However, this has no actual effect until a respawn packet is received,

The process is

Server:  Teleport Event and seed for new world sent to client
Client: newSeed variable updated, but doesn't do anything

Server: Teleport process executed, which includes sending of a respawn packet
Client: new seed is used for initialising the new WorldClient object instead of the seed of the current world.

If the teleport event is cancelled, then no harm done.

I tested this by editing the NBT file so that it was always raining.
- Log out in world1
- Log in to world1
- teleport to edge of desert in world2

Without the modification, there is rain in the desert in world 2.  With the modifications, the rain only rains on the grass side.

Two new MCP files are required, WorldInfo and NetClientHandler.  Changes use // BukkitContrib start/end

### DIFF
--- a/MCP src/NetClientHandler.java
+++ b/MCP src/NetClientHandler.java
@@ -1,0 +1,729 @@
+package net.minecraft.src;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.InetAddress;
+import java.net.Socket;
+import java.net.URL;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.Random;
+import net.minecraft.client.Minecraft;
+import net.minecraft.src.Block;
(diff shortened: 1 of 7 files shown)


Afforess:
Please update it to use tabs.

Raphfrk:
Done, I think.  I opened the file in Eclipse and used ctrl-shift-F.

It's like the opposite of Bukkit :), tabs were banned there.

Afforess:
I hate hate hate spaces. Plus, being anti-Bukkit is a good thing in some respects.

PR #44, opened 2011-07-19
Adds a basic implementation of outgoing packet listeners

desmin88:
This adds a basic implementation packet listeners listening on a per-packet type basis, to mitigate any effects of lag or slowdown caused by listeners. 

Packet Listeners are made by extending PacketListener and adding your class to the list of listeners by calling 
PacketListenerHandler.addListener().

Packet listeners are registered by the ID of packet i.e packet.b(); or Packet18, 18 being the packet id.

It trys to run a minimalistic amount of code per each packet and, quickly returns true if no registered packet listeners are found, so only few lines more of code are ran every sendPacket(); so to reduce lag effects.

I've followed the bukkit conventions of 4 spaces, not tabs and added javadocs commenting for it. My commits totaling 7 were messed up where trying to rebase, but bear with me.

Hope you merge!


Afforess:
Couple of formatting issues:

1.)

<pre><code>        //One of them said false, so, we son't send it.
        else {
            return;
        }</code></pre>

That is dead code. If it's an else case, do nothing. Just remove those 4 lines and get on with it.

2.) Fix the formatting of the pom.xml - please don't change random unrelated files.
3.) Switch to tabs. BukkitContrib uses tabs, not spaces. For more details, see the <a href="https://github.com/Afforess/BukkitContrib/blob/master/README.md">readme</a>.

desmin88:
Ahh, sorry.

The pom.xml was a mishap from when it wasn't set to compile with 1.6

Expect a new commit very soon, to fix the aformentioned issues.

Afforess:
Great. I will test this this evening and let you know my verdict.

desmin88:
Should be fixed now. The pom.xml I downloaded raw from your original repo, and formatted everything back to tabs, removed dead code. I overlooked that in the beginning.

Afforess:
You didn't push your changes to the ContribNetServerHandler. ;)

desmin88:
Jesus christ what did I do. I'm new to this git thing. Hold on.

I accidentally used Eclipse auto format, messed it all up. should be fixed now, sorry.

Afforess:
Looks good. I will take a look at it and profile it when I get home, ~6pm EST.

desmin88:
I await that time then.

desmin88:
Any news on your testing?

Afforess:
Only started looking - I just finished handling the complaints on the forums. >.>

desmin88:
Ahh, the 'perms' exception

Afforess:
Pulled for testing.

desmin88:
Sweet.

Raphfrk:
Efficiency of this pull could be improved by changing the lookup to an array.  Notch (ug) doesn't actually store the packetId within the packet, so a hash lookup for that is required as part of the Packet.b() call.

Also, I think you have to many method calls on the critical section.

The key is that when there is no listener, sendPacket should have absolutely minimum overhead.

For example, you could swap

```
public void sendPacket(Packet packet) {
    if (PacketListenerHandler.checkPacket(packet)) {
        super.a(packet);
    }
}
```

to

```
public void sendPacket(Packet packet) {
    ArrayList<PacketListener> listeners = PacketListenerHandler.listeners[packet.b()];
    if(listeners != null) {
        packet = PacketListenerHandler.processPacket(listeners, packet);
    }

    if(packet != null) {
        super.sendPacket(packet);
    }
}
```

(Also, you should use super.sendPacket() so it used the CraftBukkit sendPacket system)

PacketHandler.listeners would be

ArrayList<PacketListener> listeners[] = new ArrayList<PacketListener>[256];

That would store the listener array for each packet id.

You should also allow chaining of packets.  At the moment, all a plugin can do is cancel a packet.

```
public boolean outGoingPacket(Packet packet, int packetId);
```

could be changed to

```
public Packet outGoingPacket(Packet packet, int packetId);
```

This would allow a plugin to change the packet for another one.  In fact, maybe an ArrayList<Packet> would be better, that way extra packets could be added.  The packet(s) output from one listener would then be sent to the next listener.

Also, you could add listener priorities.

Raphfrk:
You may also require some kind of locking for PacketListenerHandler.listeners.

PR #45, opened 2011-07-19
Teleport smoothing

Raphfrk:
Mostly implements the plan from the other "thread".

A HashSet is used to store all active chunks.  Active chunks are chunks that the client has initialized.
# Server to Client Packets
## Packet51MapChunk
- does a check to make sure it is initialized
  - calls packet50PreChunk method to initialize, so activeChunks hashSet is updated
## Packet50PreChunk
- won't uninitialize if within 2 chunks of a player
- stores uninitialized chunks in unload queue
- if a packet is later initialized, removes it from the unload queue
## Packet11PlayerPosition, Packet13PlayerLookMove

This is a teleport command if sent by the server.
- Checks if target is an unloaded chunk and fast sends a 5x5 grid of chunks
  - This might be overkill
## Packet9Respawn
- Resets the activeChunks hashSet
# Client to Server Packets
## Packet10Flying
- Updates the player's chunk position if it changes

### DIFF
--- a/src/org/bukkitcontrib/ContribNetServerHandler.java
+++ b/src/org/bukkitcontrib/ContribNetServerHandler.java
@@ -4,7 +4,12 @@ import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.Collections;
+import java.util.Iterator;
 
 import org.bukkit.Bukkit;
 import org.bukkit.Location;
@@ -35,6 +40,14 @@ import net.minecraft.server.ItemStack;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.NetServerHandler;
 import net.minecraft.server.NetworkManager;
+import net.minecraft.server.ChunkCoordIntPair;
+import net.minecraft.server.Packet;
+import net.minecraft.server.Packet9Respawn;
+import net.minecraft.server.Packet10Flying;
+import net.minecraft.server.Packet11PlayerPosition;
+import net.minecraft.server.Packet13PlayerLookMove;
+import net.minecraft.server.Packet50PreChunk;
+import net.minecraft.server.Packet51MapChunk;
 import net.minecraft.server.Packet100OpenWindow;
 import net.minecraft.server.Packet101CloseWindow;
 import net.minecraft.server.Packet102WindowClick;
@@ -53,6 +66,8 @@ public class ContribNetServerHandler extends NetServerHandler{
 	protected boolean activeInventory = false;
 	protected Location activeLocation = null;
 	protected ItemStack lastOverrideDisplayStack = null;
+
+	private final int teleportZoneSize = 3; // grid size is a square of chunks with an edge of (2*teleportZoneSize - 1)
 
 	public ContribNetServerHandler(MinecraftServer minecraftserver, NetworkManager networkmanager, EntityPlayer entityplayer) {
 		super(minecraftserver, networkmanager, entityplayer);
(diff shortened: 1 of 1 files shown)


Afforess:
Pulled for testing.

Afforess:
Good news, it does work as intended, teleports do not cause the ground falling away anymore.

While I am abusing your knowledge, Any chance we can implement the Packet51MapChunkThread?

Raphfrk:
The only issue I can think of is how it will react with limited bandwidth and CPU (testing on a good computer where the server is on localhost is probably not the best test.).  It might be worth having the teleportZoneSize as a parameter.  If set to 1, it will only send the chunk the player is teleporting to and that wouldn't cause a spike any more than normal.

OTOH, it is only fast generates 25 chunks out of the 441 that need to be generated and could be reduced to 9 if a 3x3 grid was used instead.

On the Packet51 thread thing, I was thinking that too.  However, it isn't a post processing thing like the teleport change.

Zeerix had to change everywhere in the code where Packet51 is generated, which isn't to many places.
- EntityPlayer -> new chunks when you move
- PlayerInstance -> chunk portions due to block changes

I don't think you currently override PlayerInstance (and that one is a crazy name)?  It should be called chunkUpdateManager or something.  There is one per chunk, not one per player.

Playerinstance sends packet51 packets due to block changes.  If more than 10 blocks are modified in a chunk in a single clock tick, then it sends them using packet51 instead of the multi-block update.  When a block is changed, it remembers the max x, y, z and min x, y, z and it just sends the cuboid.  This means that if you changed the 8 corners of a chunk and any three other blocks, it would send the entire chunk, instead of just 11 updates :).

If that wasn't included, then chunk updates due to teleporting would still be handled by the other thread.

There is also a potential problem if Zeerix's changes are actually added.  The 2 systems could conflict and I think the Bukkit team are at least considering it.  It is the kind of thing that would be accepted.

The ideal would be to modify Packet51MapChunk so that it doesn't compress at all.  That way, it wouldn't matter where it is called from and all the changes could happen in NetServerHandler.

With the talk of class loaders yesterday, I had a look.  The plugin loader for every Plugin has CraftBukkit's class loader defined as a parent.  This means that CraftBukkit classes are loaded by the standard loading process, and all plugins get access to them, and there is no way to replace one class with another.

Btw, you "unpulled" desmin88's changes, since they mod the same file.  He also overrides sendPacket().

I think my changes should look at the packets after any changes made by others, since they supposed to track what chunks are initialized by the client.  

Maybe rename my sendPacket method to sendPacket2 and have his sendPacket method call my sendPacket2 method instead of super.sendPacket.

Btw2, seems Notch is considering a map height change

https://github.com/Bukkit/CraftBukkit/commit/d1a866e468c47a449b442c0729123081d5c8beb2
http://www.minecraftforum.net/news/145-18-updates-taller-worlds-delicious-foods/

"Additionally - Notch appears to have been experimenting with map-height changes. He went on to say that while nothing definitive is planned, they are looking into possible map height restrictions in the future."

The image, but you probably can't see it:
http://media.minecraftforum.net/content/images/2011/06/18heightmap.png

Raphfrk:
Another thought, if desmin88's changes had been in place (including detecting upstream packets), the teleport smoothing thing could have been done as a plugin :).

Raphfrk:
Found 2 links to info on plugins/security managers.

http://stackoverflow.com/questions/3947558/java-security-sandboxing-plugins-loaded-via-urlclassloader

http://stackoverflow.com/questions/502218/sandbox-against-malicious-code-in-a-java-application

Looks like the recommendation is to use another thread for potentially malicious code.  This would allow a timeout to be set for methods calls and then just kill the plugin's thread.

The process is
- save old security manager
- replace security manager with a restrictive custom one
- call sandboxMode(true) on custom security manager
- run plugin method
- call sandboxMode(false) on custom security manager
- replace security manager with original one

Since the plugin doesn't have access to the custom security manager object, it can't switch off sandbox mode.  It isn't clear if System.setSecurityManager is global ... seems to be.

That could be coupled with a class loader that has a whitelist of classes that can be used.  Reflection would be a good one not to include on the white list :).

Afforess:
I'm already aware of SecurityManager (in fact, that is the plan). I/O Access, Internet Access and reflection will all be locked out. 

I didn't accidentally unpull desmin88's changes. They were too slow. With 1 player online, no listeners, and just timing the difference in ms for the if statement, I was using an extra 50 ms/minute of execution. That's far too slow. There's just way to much volume on sendPacket(). 

Are you sure PlayerInstance is once per chunk? It looks like 1 per player on each world.

Raphfrk:
Yeah, pretty sure.

Maybe you are mixing it up with PlayerManager?

Look at the local variables

```
private List b;
private int chunkX;
private int chunkZ;
private ChunkCoordIntPair location;
private short[] dirtyBlocks;
private int dirtyCount;
private int h;
private int i;
private int j;
private int k;
private int l;
private int m;
```

It stores the chunk position (which could be set to finals).

this.b is a List with all the players in it.

It basically manages block updates for 1 chunk.

Afforess:
PlayerManager is 1 per world.

But yeah, looks like you are right: https://github.com/Bukkit/CraftBukkit/blob/master/src/main/java/net/minecraft/server/PlayerManager.java#L44

Afforess:
Looking at PlayerManager & PlayerInstance, I think we could extend and replace the PlayerManagers in each world (and check when new worlds are loaded), it's even public. It might be a bit of a hassle to move the data from the old PlayerInstances & PlayerManager to new ones, but it's do-able. In which case, the Chunk thread could be added.

Raphfrk:
Well, with enough hassle, anything is possible.  Hmm, might be worth asking Zeerix if we can use his file.  It is (presumably) under GPL though.

On the max height thing, that would so have been a potential killer-app for BukkitContrib :) (like Bukkit and many worlds), even 2x height would be cool.  OTOH, it would mean that only enabled clients could join.

_Edit By Afforess: Damn accidently edited your comment w/o knowing. Why does Github let me edit your comments? Seems like I could abuse it._

Afforess:
> Well, with enough hassle, anything is possible.  Hmm, might be worth asking Zeerix if we can use his file.  It is (presumably) under GPL though.

CraftBukkit is not licensed properly, at least according to EvilSeph. If the license is invalid, then it's public domain. But it's murky territory, I agree. No harm in rewriting it ourselves, since we have to anyway (can't edit packet fields to make them public...have to use reflection).

> On the max height thing, that would so have been a potential killer-app for BukkitContrib :) (like Bukkit and many worlds), even 2x height would be cool.  OTOH, it would mean that only enabled clients could join.

Notch says it is SP only, no chance of it being for MP, so it still can be a killer feature. I agree, 512 height is epic. Servers would need to use new maps though, and yes, clients would need to have it enabled. Speaking of which, I need a faster way to authenticate the client, because if I start adding custom blocks, etc, the new id's will be in the initial map chunks, and standard client s will choke on them. The client is authenticated fairly quickly (usually within 300ms), but I will need something even faster.

I was thinking of using the seed and world values in the Login packet to write the client version, but there is no way for me to intercept the login packet on the plugin side, it's handled in the NetLoginHandler, which I don't have access to. Do you have any good ideas on how I can authenticate before sending map and entity packets? Or possibly "clean" packets of invalid references? Until they are authenticated?

Raphfrk:
Ahh, I think if the license is invalid, then it is technically unlicensed.  Copyright defaults to "no-use" not public domain.

However, as you say murky, since he is submitting the files to a nominally GPL project.

Anyway, I was thinking that his thread handler class could probably be used "as is" ((well as you say with reflection) and then just add the hooks.  

Except for the player instance hook (which I don't think it likely a big one).

Actually, he writes directly to the network manager, so his changes would break teleport smoothing (if added to CraftBukkit), which is annoying.

In fact, any low priority packet would effective skip NetServerHandler.

Raphfrk:
Hmm, actually maybe it won't since I redesignate the "fast" Packet51 packets as fast updates.

Afforess:
> Copyright defaults to "no-use" not public domain.
> I'm not an expert, it's why I'd avoid it. I expect if we asked Zeerix he'd give us permission. I have no problems attributing his work.
> 
> Except for the player instance hook (which I don't think it likely a big one). 

If it turns out to be problematic, I can always get access. It will just be a PITA.

> Actually, he writes directly to the network manager, so his changes would break teleport smoothing (if added to CraftBukkit), which is annoying.

That's why I was hoping we could implement it. That way we get the benefit of multithreading, as well as your teleport smoothing. (Which, BTW, works wonders for teleporting.)

Now that I think about it - is there any way to smooth fast vehicles (Minecarts)? Riding in them causes you to out-race the server's packet updating after about 45 seconds of maximum speed ride time.

Raphfrk:
Hmm, if both are overridden, then his method would never be called, so the bypass would be avoided.

All low priority packets are sent to his queue.

As for outrunning chunks.  It depends on why it is happening.

When a player moves out of a chunk, it unloads chunks behind and is supposed to load chunks in front.  I think that happens quickly (the unload/init of chunks).  The player is probably considered to have those chunks loaded, and the map data is queued.

In PlayerEntity, chunkCoordIntPairQueue stores all chunks which have been initialised.  It is supposed to send them to the player.

What is probably be happening is that it is filling up faster than it can empty.  Each time you change chunks, a line of 21 chunks behind is unloaded and a line of 21 ahead is initialized and added to the queue.

If the size of array exceeds 21 \* 10 = 210, then half of the 21x21 grid would be missing.

A test could be made that if the number of chunks to send exceeds 100, then the nearest chunk to the player would be sent instead.  This would require changing the ArrayList to a LinkedHashSet().

Also, chunks which are more than a say 50 chunks from the player could be dropped, since they are outside the 21x21 chunk grid.  Actually, what is the max view distance?

Afforess:
> What is probably be happening is that it is filling up faster than it can empty. Each time you change chunks, a line of 21 chunks behind is unloaded and a line of 21 ahead is initialized and added to the queue.

Yeah, that is not ideal, especially if the player only ends up spending 16 ticks (1 tick per block) in the chunk. 

Perhaps motion could also play a role in prioritizing chunk loading, so that server does a "best guess" of where the player will be and loads chunks ahead of them, so if you have a straight rail, at least the rails load.

> Also, chunks which are more than a say 50 chunks from the player could be dropped, since they are outside the 21x21 chunk grid. Actually, what is the max view distance?

15, but no one uses that. 10 is default, but 3 is minimum. 5-7 is common.

Raphfrk:
I think just loading from nearest to furthest would be fine (esp if dist = dx + dy is used).  The problem is chunks that are 10 blocks away at 90 degrees to travel have higher priority than the one directly ahead.

Scanning 100 blocks for the nearest would be a bit of an overhead though.  However, chunk loads don't happen that often, and have CPU load anyway, so just scanning for the nearest one wouldn't be that big a deal.

Raphfrk:
Hmm, PlayerInstance is not a public class.  Is is possible to use reflection (or something similar) to extend it?

Raphfrk:
Also, for extending classes:

This will basically transfer a named field from one object to another.

transferField(manager, this, "a");

will move the "a" field to this object.

```
    private static void transferField(Object src, Object dest, String fieldName) {
            try {
                    Field field = src.getClass().getDeclaredField(fieldName);
                    field.setAccessible(true);
                    Object temp = field.get(src);
                    field.set(dest, temp);
            } catch (NoSuchFieldException e) {
                    throw new RuntimeException(e);
            } catch (IllegalAccessException e) {
                    throw new RuntimeException(e);
            }
    }
```

The basic ContribPlayerManager can be written as

```
    package org.bukkitcontrib;

    import net.minecraft.server.MinecraftServer;
    import net.minecraft.server.WorldServer;
    import net.minecraft.server.PlayerList;
    //import net.minecraft.server.PlayerInstance;
    import net.minecraft.server.PlayerManager;

    import java.util.List;
    import java.lang.reflect.Field;

    public class ContribPlayerManager extends PlayerManager {

            public static void replacePlayerManager(WorldServer world) {
                    if(!(world.manager instanceof ContribPlayerManager)) {
                            world.manager = new ContribPlayerManager(world.manager);
                            System.out.println("Replaced " + world.worldData.name);
    //                      ContribPlayerInstance.replacePlayerInstances(world.manager);
                    }
            }

            public ContribPlayerManager(PlayerManager manager) {
                    super(null, 0, 10);
                    transferField(manager, this, "a");
                    transferField(manager, this, "b");
                    transferField(manager, this, "server");
                    transferField(manager, this, "d");
                    transferField(manager, this, "e");
                    transferField(manager, this, "f");
            }

            private static void transferField(Object src, Object dest, String fieldName) {
                    try {
                            Field field = src.getClass().getDeclaredField(fieldName);
                            field.setAccessible(true);
                            Object temp = field.get(src);
                            field.set(dest, temp);
                    } catch (NoSuchFieldException e) {
                            throw new RuntimeException(e);
                    } catch (IllegalAccessException e) {
                            throw new RuntimeException(e);
                    }
            }

    }
```

which is a lot simpler than doing reflection for each field.

Afforess:
Does it matter? For control of PlayerInstances, we need to replace each worlds PlayerManager. PlayerManager is the only place I saw even create PlayerInstances. If we can rewrite PlayerManager then we could rewrite PlayerInstances to a new class, so who cares?

Raphfrk:
True, but it means making a full copy rather than just overriding the one method that created a Packet51

Afforess:
> True, but it means making a full copy rather than just overriding the one method that created a Packet51

Sadly, I see no other choice. At least we get to use coding conventions and unobfuscated names.

Afforess:
One other advantage I saw to controlling the player manager I saw was that we could allow plugins to change the view distance on a player by player basis instead of a server wide setting. E.g admins see all 15 chunks, regulars see 5.

Raphfrk:
It would also be a way for admins to increase donations :).  Guests get a 3 chunk view distances.

Afforess:
> It would also be a way for admins to increase donations :). Guests get a 3 chunk view distances.

Exactly. Am also tempted to code in special consideration for myself, but I will probably resist the urge. ;) Can I assume you are working on it? 

I've got textures with the GUI 95% done, just need to clean the code up and push it. Then I can actually start moving on the other parts. Textures were a PITA because I had to implement every step of them...Instead of using existing code.

Edit: There is a reason for the 15 chunk limit - I did a test and increased it on the client, and past 16, I got out of memory errors pretty fast. I expect very few machines can handle that. But as for the 3 bottom limit, really, you only need 1 chunk, the one you are on...heh.

Raphfrk:
- Exactly. Am also tempted to code in special consideration for myself, but I will probably resist the urge. ;) Can I assume you are working on it? 

There was that mod who gave himself admin permissions on any server which had installed the mod and caused a reaction.  I think he kicked an admin on a server he logged into :).  

Ofc, that isn't the same as giving yourself a larger view distance.

You need a 3x3 chunk grid, since it only checks if you have moved every 10 blocks.  With only 1 chunk you could walk off the edge.  That works out as a view distance of 1.

I have a version that compiles and doesn't seem to do anything crazy.  However, it doesn't do replacement back to normal PlayerInstances on disable.  In fact, since PlayerInstance is private, it wouldn't be possible to create PlayerInstance objects from ContribPlayerInstance objects.

However, it doesn't cause exceptions on reload, I guess the replaced files stay replaced :).

Ofc, it is just a copy of the standard classes and doesn't have any new features.

Afforess:
> There was that mod who gave himself admin permissions on any server which had installed the mod and caused a reaction. I think he kicked an admin on a server he logged into :).
> 
> Ofc, that isn't the same as giving yourself a larger view distance.

Yeah. I am familiar with that incident. There are other examples of such abuse. It's part of the reason why I will resist the urge, despite the fact that I would enjoy the 15 chunk view.

> I have a version that compiles and doesn't seem to do anything crazy. However, it doesn't do replacement back to normal PlayerInstances on disable. In fact, since PlayerInstance is private, it wouldn't be possible to create PlayerInstance objects from ContribPlayerInstance objects.

Well yeah, just because they were created by a different class loader doesn't mean they don't work. As long as your custom PlayerManager expects them, they will work fine when the plugin is disabled. And when It's re-enabled, we'll just reset them all over again, assuming the field mappings are the same, or you added a good way to reset them.

Cool though.

Afforess:
You can quote text with the right arrow symbol, BTW.

Raphfrk:
> You can quote text with right arrow symbol, BTW

Thanks.

I think I am just going to use Zeerix's code, it only needs tiny changes to be compatible.  I sent him a PM on the forums, but will just integrate it anyway.  I doubt he will object, we can always recode it, if he does.

Raphfrk:
Hmm, you don't override EntityPlayer?

Raphfrk:
Hmm, lots of references to EntityPlayers.  It might be easier to try to intercept the method call.

However, the method is the clockTick method that is sent to all entities.

Afforess:
> Hmm, you don't override EntityPlayer?

That's a bit more work than I ever intended. I guess i'd do it if it came down to it, but no, I don't.

Raphfrk:
Was talking to Zeerix and he said that it is fine.

Turns out it isn't necessary to replace EntityPlayer.

All chunk load/unload stuff is handled when the move packet Packet10 is received.  I can just override that one method in NetServerHandler and wipe the chunk load queue before it even gets to EntityPlayer.

This means that EntityPlayer won't actually send chunks, since the queue will be wiped.

I can then manually send them from NetServerHandler.

Afforess:
Sweet.

PR #47, opened 2011-07-21
Port of Zeerix's MapChunkThread system

Raphfrk:
This has almost no testing.  

I teleported between 2 distance places and back and nothing exploded.

### DIFF
--- a/src/org/bukkitcontrib/BukkitContrib.java
+++ b/src/org/bukkitcontrib/BukkitContrib.java
@@ -9,14 +9,17 @@ import java.net.URL;
 import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.util.logging.Logger;
+import java.util.List;
 
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.entity.Player;
+import org.bukkit.World;
 import org.bukkit.event.Event.Priority;
 import org.bukkit.event.Event.Type;
 import org.bukkit.plugin.java.JavaPlugin;
 import org.bukkit.util.FileUtil;
+import org.bukkit.craftbukkit.CraftWorld;
 import org.bukkitcontrib.block.ContribCraftChunk;
 import org.bukkitcontrib.config.ConfigReader;
 import org.bukkitcontrib.inventory.ItemManager;
@@ -111,9 +114,16 @@ public class BukkitContrib extends JavaPlugin{
 			sendBukkitContribVersionChat(player);
 			playerListener.manager.onPlayerJoin(player);
 		}
+
+		List<World> worlds = getServer().getWorlds();
+		for (World world : worlds) {
+			ContribPlayerManager.replacePlayerManager(((CraftWorld)world).getHandle());
+		}
 		
 		ContribCraftChunk.replaceAllBukkitChunks();
 		appearanceManager.onPluginEnable();
+
+		MapChunkThread.startThread(); // Always on
 		
 		//Start counting ticks
 		Bukkit.getServer().getScheduler().scheduleSyncRepeatingTask(this, new ServerTickTask(), 0, 1);
--- a/src/org/bukkitcontrib/ContribNetServerHandler.java
+++ b/src/org/bukkitcontrib/ContribNetServerHandler.java
@@ -560,7 +560,7 @@ public class ContribNetServerHandler extends NetServerHandler{
 		return currentChunk.get();
 	}
 
-	private void setPlayerChunk(int cx, int cz) {
+	public void setPlayerChunk(int cx, int cz) {
 		ChunkCoordIntPair cur = currentChunk.get();
 		if(cur.x != cx || cur.z != cz) {
 			currentChunk.set(new ChunkCoordIntPair(cx, cz));
--- a/src/org/bukkitcontrib/ContribNetServerHandler.java
+++ b/src/org/bukkitcontrib/ContribNetServerHandler.java
@@ -10,12 +10,16 @@ import java.util.Map;
 import java.util.Set;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.bukkit.Bukkit;
 import org.bukkit.Location;
 import org.bukkit.craftbukkit.entity.CraftPlayer;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Event.Result;
+import org.bukkit.event.player.PlayerMoveEvent;
 import org.bukkitcontrib.event.inventory.InventoryClickEvent;
 import org.bukkitcontrib.event.inventory.InventoryCloseEvent;
 import org.bukkitcontrib.event.inventory.InventoryCraftEvent;
@@ -60,6 +64,7 @@ import net.minecraft.server.Packet106Transaction;
 import net.minecraft.server.Slot;
 import net.minecraft.server.TileEntityDispenser;
 import net.minecraft.server.TileEntityFurnace;
+import net.minecraft.server.AxisAlignedBB;
 
 public class ContribNetServerHandler extends NetServerHandler{
 	protected Map<Integer, Short> n = new HashMap<Integer, Short>();
@@ -386,22 +391,32 @@ public class ContribNetServerHandler extends NetServerHandler{
 
 	@Override
 	public void sendPacket(Packet packet) {
+		if(packet != null) {
+			if(packet.k) {
+				MapChunkThread.sendPacket(this.player, packet);
+			} else {
+				sendPacket2(packet);
+			}
+		}
+	}
+
+	public void sendPacket2(Packet packet) {
 		if(packet instanceof Packet51MapChunk) {
-			sendPacket((Packet51MapChunk)packet);
+			sendPacket2((Packet51MapChunk)packet);
 		} else if(packet instanceof Packet50PreChunk) {
-			sendPacket((Packet50PreChunk)packet);
+			sendPacket2((Packet50PreChunk)packet);
 		} else if(packet instanceof Packet11PlayerPosition) {
-			sendPacket((Packet11PlayerPosition)packet);
+			sendPacket2((Packet11PlayerPosition)packet);
 		} else if(packet instanceof Packet13PlayerLookMove) {
-			sendPacket((Packet13PlayerLookMove)packet);
+			sendPacket2((Packet13PlayerLookMove)packet);
 		} else if(packet instanceof Packet9Respawn) {
-			sendPacket((Packet9Respawn)packet);
+			sendPacket2((Packet9Respawn)packet);
 		} else {
 			super.sendPacket(packet);
 		}
 	}
 
-	public void sendPacket(Packet50PreChunk packet) {
+	public void sendPacket2(Packet50PreChunk packet) {
 		int cx = packet.a;
 		int cz = packet.b;
 		boolean init = packet.c;
@@ -443,26 +458,26 @@ public class ContribNetServerHandler extends NetServerHandler{
 		}
 	}
 
-	public void sendPacket(Packet51MapChunk packet) {
+	public void sendPacket2(Packet51MapChunk packet) {
 		ChunkCoordIntPair chunkPos = new ChunkCoordIntPair(packet.a >> 4, packet.c >> 4);
 		if(!activeChunks.contains(chunkPos)) {
-			//System.out.println("Packet51: Sending emergency initialize for " + chunkPos.x + " " + chunkPos.z);
-			sendPacket(new Packet50PreChunk(chunkPos.x, chunkPos.z, true));
+			//System.out.println("Packet51: dropping data packet being sent to uninitialized chunk " + chunkPos.x + " " + chunkPos.z);
+			return;
 		}
 		super.sendPacket(packet);
 	}
 
-	public void sendPacket(Packet11PlayerPosition packet) {
+	public void sendPacket2(Packet11PlayerPosition packet) {
 		playerTeleported(((int)packet.x) >> 4, ((int)packet.z) >> 4);
 		super.sendPacket(packet);
 	}
 
-	public void sendPacket(Packet13PlayerLookMove packet) {
+	public void sendPacket2(Packet13PlayerLookMove packet) {
 		playerTeleported(((int)packet.x) >> 4, ((int)packet.z) >> 4);
 		super.sendPacket(packet);
 	}
 
-	public void sendPacket(Packet9Respawn packet) {
+	public void sendPacket2(Packet9Respawn packet) {
 		activeChunks.clear();
 		super.sendPacket(packet);
 	}
@@ -472,10 +487,55 @@ public class ContribNetServerHandler extends NetServerHandler{
 		if(packet instanceof Packet11PlayerPosition || packet instanceof Packet13PlayerLookMove) {
 			setPlayerChunk(((int)packet.x) >> 4, ((int)packet.z) >> 4);
 		}
+		manageChunkQueue(true);
 		super.a(packet);
 	}
 
-	private ChunkCoordIntPair currentChunk = new ChunkCoordIntPair(Integer.MAX_VALUE, Integer.MIN_VALUE);
+	private final LinkedHashSet<ChunkCoordIntPair> chunkUpdateQueue = new LinkedHashSet<ChunkCoordIntPair>();
+
+	private final AtomicInteger updateCounter = new AtomicInteger();
+
+	private final int[] spiralx = new int[] {0, -1, -1, -1,  0,  1,  1,  1,  0, -2, -2, -2, -2, -2, -1,  0,  1,  2,  2,  2,  2,  2,   1,  0, -1};
+	private final int[] spiralz = new int[] {0, -1,  0,  1,  1,  1,  0, -1, -1, -2, -1,  0,  1,  2,  2,  2,  2,  2,  1,  0, -1, -2,  -2, -2, -2};
+
+	// This may not catch 100% of packets, but should get most of them, a small number may end up being compressed by main thread
+	public void manageChunkQueue(boolean flag) {
+		List<ChunkCoordIntPair> playerChunkQueue = player.chunkCoordIntPairQueue;
+
+		if(!playerChunkQueue.isEmpty()) {
+			Iterator<ChunkCoordIntPair> i =  playerChunkQueue.iterator();
+			while(i.hasNext()) {
+				ChunkCoordIntPair next = i.next();
+				chunkUpdateQueue.add(next);
+			} 
+			playerChunkQueue.clear();
+		}
+
+                if(!chunkUpdateQueue.isEmpty() && (b() + MapChunkThread.getQueueLength(this.player)) < 4) {
+			ChunkCoordIntPair playerChunk = getPlayerChunk();
+			ChunkCoordIntPair first = chunkUpdateQueue.iterator().next();
+			if(updateCounter.get() > 0) {
+				int cx = playerChunk.x;
+				int cz = playerChunk.z;
+				boolean chunkFound = false;
+				for(int c = 0; c < spiralx.length; c++) {
+					ChunkCoordIntPair testChunk = new ChunkCoordIntPair(spiralx[c] + cx, spiralz[c] + cz);
+					if(chunkUpdateQueue.contains(testChunk)) {
+						first = testChunk;
+						chunkFound = true;
+						break;
+					}
+				}
+				if(!chunkFound) {
+					updateCounter.decrementAndGet();
+				}
+			}
+			chunkUpdateQueue.remove(first);
+                       	MapChunkThread.sendPacketMapChunk(first, this.player, this.player.world);
+                }
+	}
+
+	private final AtomicReference<ChunkCoordIntPair> currentChunk =  new AtomicReference<ChunkCoordIntPair>(new ChunkCoordIntPair(Integer.MAX_VALUE, Integer.MIN_VALUE));
 
 	private final Set<ChunkCoordIntPair> activeChunks = Collections.synchronizedSet(new HashSet<ChunkCoordIntPair>());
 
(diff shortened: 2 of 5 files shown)


Afforess:
> This has almost no testing.

My kind of pull request.

> I teleported between 2 distance places and back and nothing exploded.

Awesome.

Afforess:
I will test this out tomorrow evening and over the weekend and let you know. ;)

Raphfrk:
> My kind of pull request

:), was late and figured might as well send what I had.

I might make a few tweaks this evening to the chunk load ordering.  

The new queue uses a LinkedHashSet instead of a LinkedList.  This means it is fast to check if the nearest 3x3 grid of chunks are in the queue.  Moving queue management to NetServerHandler, makes it much easier to tweak it.

In fact, I could have a chunkChanged flag that gets set when the player changes chunk, since I am detecting chunk changes anyway.  That way it would only check if all nearby chunks are loaded once every time the player changes chunk.

Also, it needs to react to world loading events.  At the moment, it only replaces worlds that are loaded when the plugin is enabled.

Zeerix's pull also has a call to b() in NetServerHandler to tell how many low priority packets are in the queue.  I missed that.

I was re-reading the smoothing pull and I wonder if "emergency" chunk initialization is a bad idea.  This happens when the server tries to send a packet51, but the target chunk is not active.  

This could cause a memory leak.  In fact, if anything, those packets should not be sent at all, since they are a waste of bandwidth, since the client ignores them.  Also, when prioritizing the nearby packets, it is possible that there would be loads of chunks waiting to be sent in the queue.  If they cause an initialization when sent, they would never clear.

So, I remember the list:
- Prioritize the 5x5 grid of chunks nearest the player (but only check after chunk changes)
- Instead of sending the first chunk on the queue, send the first chunk that is within 20 chunks of the player
- Chunks more than 20 from the player are considered stale and dropped instead of sending the data to the client
- Swap the PlayerManager for worlds loaded while the plugin is active
- Add call to b() for making sure not to flood the output buffer
- Remove emergency chunk initialize and drop packets instead
- Teleport smoothing should manually initialize the target chunks

Raphfrk:
Hmm, seems you can create instances of private classes :)

<b>LoadTst.java</b>

```
import java.lang.reflect.Method;
import java.lang.reflect.Constructor;

public class LoadTst {

    public static void main(String[] args) throws Exception {

        Object o;
        Class clazz = LoadTst.class.getClassLoader().loadClass("priv.priv.ReadTst");
        Constructor c = clazz.getDeclaredConstructor(new Class[] {});
        c.setAccessible(true);
        o = c.newInstance(new Object[] {});
        Method m = clazz.getDeclaredMethod("main", new Class[] {String[].class});
        m.setAccessible(true);
        m.invoke(o, new Object[] {new String[] {}});

    }

}
```

<b>ReadTst.java</b>

```
package priv.priv;

class ReadTst {

    private void main(String[] args) {

        System.out.println("Starting");

    }

}
```

Afforess:
> Hmm, seems you can create instances of private classes :)

Now that's cool.

I <3 java.

Raphfrk:
In fact, it should be possible to write a method of the form

Object cloneObject(Object src, Class dest);

This would create an object of type dest and copy all the fields from src to dest.

Raphfrk:
I think it still wouldn't allow extension of the PlayerInstance class though.  However, it would allow converting ContribPlayerInstances back to PlayerInstances.

Raphfrk:
Btw, apparently, the problem with AuthMe is that it blocks all command until the user has entered their password.

In theory, Bukkit Contrib could automatically remember a password locally and enter it on login, saving the player the trouble.

However, dunno if you want to go down the route of making it easier for players who are using unofficial versions of the software.

Also, on your login/authentication issue.  The code already processes all Packet51 packets.  When a player logs in, you could just not send any Packet51 packets until the client has been authenticated one way or another.

With this pull, 100% of Packet51 packets (except the teleport smoothing ones) are sent to the MapChunkThread class.  It then generated the chunk and calls sendPacket2() of ContribNetServerHandler.  

Most chunks could be left in the new Queue until authentication (so Packet51s wouldn't even be generated) and the few that make it past that could be buffered.

Raphfrk:
Opps, closed by accident.  They shoudn't put comment and comment&close so close together.

Raphfrk:
Also, if you wanted to get very fancy, you could have it so that new blocks are swapped for old blocks.

This would require scanning the 81920 blocks, but that can't add much more processing than compression already does.

When registering a new block, the plugin could specify a standard block for replacement.

Afforess:
> I think it still wouldn't allow extension of the PlayerInstance class though.

Yeah, I know, I thought about that.

> However, it would allow converting ContribPlayerInstances back to PlayerInstances.

I don't think that's needed. If our custom PlayerManager remains in place, then our new PlayerInstances work just the same as the old one, without the need to convert back.

> Btw, apparently, the problem with AuthMe is that it blocks all command until the user has entered their password.
> 
> In theory, Bukkit Contrib could automatically remember a password locally and enter it on login, saving the player the trouble.
> 
> However, dunno if you want to go down the route of making it easier for players who are using unofficial versions of the software.

I started ignoring whether or not the command was cancelled in the CommandPreProcess event, I wonder how it's blocking it there. I guess I could start override handleChat in the NSH, but that seems extreme. (I wonder if it would speed up authentification though, how slow are events?)

> This would require scanning the 81920 blocks, but that can't add much more processing than compression already does.

If a plugin registers a custom block/item/entity, I will force the server to authenticate players, regardless of settings, because of the nature of the changes. Blocking the chunks until authentication seems like a decent way to do things, and should save CPU if they are kicked after failing.

Afforess:
> I might make a few tweaks this evening to the chunk load ordering. 

when you get done, let me know. I think I'll get some people to test this and the GUI stuff over the weekend. I am curious as to how big a boost to performance this provides.

Raphfrk:
> I am curious as to how big a boost to performance this provides.

Hopefully, at least compensated for any CPU spike due to smoothing :).

In theory, Zeerix's code could be modified so that it works with a thread pool instead of just 1 other thread.  It would require indexing the packets so that they are can be placed in the right order after they have been compressed.

However, that would only be required if the 2nd core is hitting 100% and does require knowing how much of a benefit the whole system is.

It would be interesting to see what are the major loads on the MC server and try to do the same thing.  Lighting updates or entity AI movement might be possibilities.

Bandwidth optimization is also a potential feature, that is what CraftProxy ended up being most used for.

With direct access to the packets (rather than having to parse the protocol stream), it would be much easier.  Also, it would eliminate having to compress and then decompress the packet data.

The way I did it was set the size field of the Packet51 packets to a minus number.  That indicated that they used the cache protocol, instead of the normal chunk format.  

Each chunk was broken up into 8 block high sections (16 blocks high for nibble data) and a hash generated for each section.  On connect, and periodically, the client would send a list of hashes that it knows about.  Also, the server would assume that the client knew about any hashes for any blocks sent during the current connection.

That gives 2048 bytes per sub-chunk and 40 hashes per chunk.  Each hash was 8 bytes long and was added to the end of the chunk data in cached mode.  When a hash was provided, the corresponding part of the chunk was set to all zeros (since the data was going to be compressed, this was easier than removing it).  For decompression, the client scanned the hashes and if it knew about the hash, it replaced the corresponding part of the chunk data.

Another possible saving is entity movement data.  CraftProxy gives bandwidth usage:

http://forums.bukkit.org/threads/inactive-tp-server-port-v376-teleport-between-servers-953.777/page-58#post-247807

Packet51 is around 70% of the bandwidth and 31  (Entity Relative move) and 33 ((Entity Relative move and look) account for most of the rest.

Afforess:
> Bandwidth optimization is also a potential feature, that is what CraftProxy ended up being most used for.

That could be a potential config setting - bandwitdth compression. I know smaller servers struggle with that, and so 50% compression would mean an extra player or two could be on.

> In theory, Zeerix's code could be modified so that it works with a thread pool instead of just 1 other thread. It would require indexing the packets so that they are can be placed in the right order after they have been compressed.

One has to weight the pros and cons. Is the second thread ever likely to even hit max usage? If we have a heavy load on MC, will we be using 100% of 2 cores now, just not one? I doubt it. More threads would be a lot of extra work w/o good justification.

> It would be interesting to see what are the major loads on the MC server and try to do the same thing. Lighting updates or entity AI movement might be possibilities.

AI would be extremely useful to move off of server ticks and onto a separate thread. On HeroCraft, a large server (#9 largest, IIRC...) where I often spend what little free playtime I have, mobs move so slowly that combat becomes trivial. Proper combat speeds would be a huge boon to big servers.

I'm acquainted with the HeroCraft server owner, so I may be able to get him to test the latest BC over the weekend - he runs the official version already. That would be extremely handy for testing this large scale.

Lighting would likely be a big boost as well.

> Each chunk was broken up into 8 block high sections (16 blocks high for nibble data) and a hash generated for each section. On connect, and periodically, the client would send a list of hashes that it knows about. Also, the server would assume that the client knew about any hashes for any blocks sent during the current connection.

Well, I have full client access, so there is no reason this can't be a fully integrated feature for BC players.

BukkitContrib opens up so many possibilities like this - I've got so many feature requests I couldn't do them all with 10 people...

Raphfrk:
> BukkitContrib opens up so many possibilities like this - I've got so many feature requests I couldn't do them all with 10 people...

Also, in theory, BukkitContrib is supposed to be an API :).

Moving entity AI to another thread would require going through the entity methods and then seeing what methods that they call.  There is a lot of stuff like calculating paths that only really needs a snap-shot of the world data.  Maybe Entities could be done in parallel, rather than moving the process to another thread exactly.

For example, if you have a for loop that calls a method a(), then you can parallelize that for loop and you only need to make sure the a() method is thread safe relative to itself.

Afforess:
> Also, in theory, BukkitContrib is supposed to be an API :).

I'd hardly call Bukkit an API either. There are just too many issues that can be fixed with vanilla to leave it alone. 

> Moving entity AI to another thread would require going through the entity methods and then seeing what methods that they call. There is a lot of stuff like calculating paths that only really needs a snap-shot of the world data. Maybe Entities could be done in parallel.

Especially now that there are chunk snapshots. How expensive is it to get a chunk snapshot? Is it worth caching them per tick? 

> For example, if you have a for loop that calls a method a(), then you can parallelize that for loop and you only need to make sure the a() method is thread safe relative to itself.

I understand. I don't think there are a lot of world calls beyond the pathfinding, it might not be too hard.

HOWEVER, how do you plan on getting control of the entities?

Raphfrk:
> HOWEVER, how do you plan on getting control of the entities?

Dunno :), was just thinking about the problem in principle.

Afforess:
New entities call the creature spawn event, so tracking that won't be a problem, but it looks like entities are stored in many places. It doesn't seem to be impossible, but it would definitely be a magnitude more difficult than the Chunk thread was.

Raphfrk:
Do you know any good plugins for moving fast, to test the changes for chunk load prioritization?

Afforess:
Minecart Mania? Seriously though, you can setup a minecart to go faster than the default speed, which is a good way to outrun chunks.

You just need core and sign commands.

http://ci.craftfire.com/view/Afforess/

http://www.wiki.afforess.com/Minecart_Mania-Plugin/Sign_Commands#Maximum_Speed_Signs

Raphfrk:
> Minecart Mania?

Hmm, but then I have to build the track :).

I don't suppose it has a command like "/mm buildtrack 1000" and it will do it for me?

Afforess:
> I don't suppose it has a command like "/mm buildtrack 1000" and it will do it for me?

WorldEdit, my friend.

Raphfrk:
Anyway, seems active redstone rails actually count as stop :p ... bah.

Afforess:
> Anyway, seems active redstone rails actually count as stop :p ... bah.

Reversed by default in MM. Seems to catch everyone off guard. Powered = Brake, Unpowered = Boost.

Raphfrk:
Hmm, what max speed do you set ?  The client crashed once ... maybe I need a longer track.

Also, what distance have you set for view distance?

Afforess:
I was using the default view distance, 10. I have tested it with maximums speeds up to 10000 (10000 / 100 = 10 x normal speed), on straight tracks.

To test stuff, I build a 12x2000 flat track area, with a oval loop around all the way to each end. 

Run the client in console mode so you can see exceptions when it crashes. (I posted how to on the BC thread if you're unfamiliar with it)

Raphfrk:
Actually, is a client crash what happens when you get near the edge ?

Afforess:
> Actually, is a client crash what happens when you get near the edge ?

No, I've been riding in a minecart, with no chunks loaded at all, just fine.

It just looks really really odd.

Raphfrk:
Hmm, is there a world edit command to create a 2000x10 rectangle of blocks :P ?

Raphfrk:
Also, the crash was an out of memory crash ... maybe I need to switch to 64 bit java

Afforess:
> Hmm, is there a world edit command to create a 2000x10 rectangle of blocks :P ? 

Definitely. Select a 10 x1 area, face the direction you want to go an additional 2k blocks in, type //expand 2000

Now you're selection area is 10x2001. Then do //set 1 (e.g to set to stone).

WorldEdit is awesome for testing out stuff.

> Also, the crash was an out of memory crash ... maybe I need to switch to 64 bit java

Possibly, but ideally, we shouldn't be crashing 32 bit clients with that, lots of people still use 32 bit Java.

Raphfrk:
Man, worldedit hates me.  By any chance would more than 1 powered rail in a row count as a brake too?

Raphfrk:
Actually, on 3rd thoughts, I think the crashes could be world related or something.

Afforess:
> Man, worldedit hates me.

It's really hard to get used to,  I struggled a lot too. Partly because it really needs a GUI.

> By any chance would more than 1 powered rail in a row count as a brake too?

Shouldn't. They are all unpowered, correct? If that's bugging you, you can change it in the config.

Afforess:
> Actually, on 3rd thoughts, I think the crashes could be world related or something.

Hmm?

Raphfrk:
Next question in the world edit tutorial :), is there a way to give items?

Raphfrk:
> Hmm

I meant maybe it is a corrupted world.

I have tried again with a brand new world

Raphfrk:
> Hmm

I meant maybe it is a corrupted world.

I have tried again with a brand new world

Afforess:
Um, WorldEdit can't give items, but Essentials or CommandBook have give commands.

With either, use

/i (itemid) (amount)

e.g

/i 1 4 

-Gives yourself 4 stone

Afforess:
> I meant maybe it is a corrupted world.

Too bad there is no way to detect and delete corrupted chunks from the server.

Raphfrk:
Hmm, I built a track that 2000 x 6 oval and the 2000 long rails are all powered rails.  

I think Minecraft Mania thinks I am still on the cart, or the game does, since when I log in with MM loaded, it starts to unload chunks.

Raphfrk:
Btw, all testing is with vanialla client so far ... so not sure what is up :p

Raphfrk:
Hmm, just to confirm an Op can place the sign ... also does it work when near powered rails, or just normal?

Afforess:
> I think Minecraft Mania thinks I am still on the cart, or the game does, since when I log in with MM loaded, it starts to unload chunks.

The game might. Since the vehicle does not get saved, restarting the server should fix it.

> Hmm, just to confirm an Op can place the sign ... also does it work when near powered rails, or just normal?

Correct.

I'd be happy to stop in and test with you - ip?

Raphfrk:
Bleh, needed to update the xml ... 

<MaximumMinecartSpeedPercent>10000</MaximumMinecartSpeedPercent>

Raphfrk:
Anyway, it happens pretty quickly now.

Afforess:
I'm on #bukkitdev if you want to chat in real time.

Raphfrk:
Hmm, when in a minecart, the client doesn't send position updates.

Raphfrk:
Found the issue:

After

```
    MapChunkThread.sendPacketMapChunk(first, this.player, this.player.world);
```

Add: 

```
    WorldServer worldserver = (WorldServer) player.world;
    List tileEntities = worldserver.getTileEntities(first.x * 16, 0, first.z * 16, first.x * 16 + 16, 128, first.z * 16 + 16);
    for(Object tileEntityObject : tileEntities) {
        if(tileEntityObject != null && tileEntityObject instanceof TileEntity) {
           TileEntity tileEntity = (TileEntity) tileEntityObject;
           Packet tilePacket = tileEntity.f();
           if(tilePacket != null) {
             MapChunkThread.sendPacket(this.player, tilePacket);
           }
        }
    }
```

That's what I get for making assumptions about what packets were sent. 

I think 

```
WorldServer worldserver = (WorldServer) player.world; 
```

is always OK.  

NetServerHandler uses

```
WorldServer worldserver = this.minecraftServer.getWorldServer(this.player.dimension);
```

However, minecraftServer is private for NetServerHandler.  I guess I could use reflection.

Raphfrk:
Actually, instead of using distance to decide if there is any point in sending a Packet51, I could just use activeChunks.

The problem is that if you were travelling fast, since all packet compresses are queued, once you stop, the server will compress all the chunks that you passed, but then won't actually send them to you.

So, needed changes:
- Restore dropping chunk compress for chunks that are far away, but use activeChunks instead.
- Add WorldLoadEvent support
- Add tile entity bug fix
- Update player chunk using PlayerMoveEvents instead of  Packet11PlayerPosition and Packet13PlayerLookMove
- Move transferField to org.bukkitcontrib.util.ReflectUtil

Afforess:
Interesting. I'm not in at work yet, just leaving. Can you set up a new pull for those?

Raphfrk:
Yeah, but can't do it now.

The only potential issue is reading player.world ... in theory, that could be updated from a different thread.  I haven't done a grep to check though.

However, NetServerHandler uses this.player.dimension in the call to getWorldServer and I can't see player.dimension changes happening on a different thread to player.world.

Raphfrk:
Btw, do you think it would be worth moving TransferField to org.bukkitcontrib.util.ReflectUtil as a static method?

I also found a bug, it won't work for PlayerInstance transfers, since it assumes the source and destination classes are compatible.

There could even be a transferAllFields() that moves all fields in the source to the target.

Afforess:
> The only potential issue is reading player.world ... in theory, that could be updated from a different thread. I haven't done a grep to check though.

I already have some code in the ContribPlayerListener that checks for teleports to a new world, so you could do a check there.

> Btw, do you think it would be worth moving TransferField to org.bukkitcontrib.util.ReflectUtil as a static method?

Yes, it sounds useful.

Raphfrk:
> I already have some code in the ContribPlayerListener that checks for teleports 

Actually, if I had checked the code, NetServerHandler.a(Packet10Flying packet10flying) calls EntityPlayer.a(boolean flag) which accesses player.world and NetServerHandler.a(Packet10Flying) is the method we are overriding.

Afforess:
> Actually, if I had checked the code, NetServerHandler.a(Packet10Flying packet10flying) calls EntityPlayer.a(boolean flag) which accesses player.world and NetServerHandler.a(Packet10Flying) is the method we are overriding.

Excellent. I'm eager to test this again this evening.

Raphfrk:
I think with the switch to player move event, it will be able to re-schedule properly for minecarts.  

Btw, even with a speed of 300, I was hitting the edge.

Afforess:
> Btw, even with a speed of 300, I was hitting the edge.

Not clear what you mean - with 300 you were outrunning chunks?

On bigger servers, you can outrun them on the vanilla speed, if you wait long enough. VoxelBox's transit lines have this problem.

> I think with the switch to player move event, it will be able to re-schedule properly for minecarts. 

Again, to be clear, this will improve the chunk loading for minecarts (specially minecarts, or any vehicle?)

Raphfrk:
> Not clear what you mean - with 300 you were outrunning chunks?

Yeah, 300% and rails made entirely of powered tracks.

> specially minecarts, or any vehicle?

Any, I think.  However, I only checked the player move event for minecarts.

Afforess:
> Any, I think. However, I only checked the player move event for minecarts.

How about checking for the generic Vehicle class instead? Eventually there will be more vehicles types, so best to get them out of the way now.

Raphfrk:
I checked it by printing to console every time the event fired, rather than looking at the code.  The console gets spammed even when in a Minecart.

Anyway, the E() method in Entity seems to copy the changes in the position of a vehicle to the position of the client.  Since the player move event tracks that player's locX and locY, then it should work for any vehicle.

Afforess:
Okay. Sounds good.

FYI, I can't reopen a merged pull request, additional changes must be in a new one.

Raphfrk:
> I can't reopen a merged pull request

That branch is updated anyway and it looks like I can submit a pull from a branch more than once.

stephenwilliams:
@Afforess and @Raphfrk I was talking to Zeerix and he mentioned this version of his MapChunkThread System: https://github.com/Rigby90/CraftBukkit/commit/c5f94cf1ef24ae4d9ef0697798f87b19aa533f21

Raphfrk:
@alta189

What's the difference, it looks very similar, except for the name of the class is changed.

Raphfrk:
Hmm, looks like he just stripped out the compression code from Packet51.  

I think the current method is better, since it would enable/disable/reload more safely.  Uncompressed Packet51 packets are generated by MapChunkThread, rather than having to change the Packet51 class.

The old PlayerInstance class would just generate a standard Packet51 packet.

stephenwilliams:
@Raphfrk
I don't know what the difference was, didn't really have time to take a look, but Zeerix was telling me that this version will have a better chance at having it pull request accepted

Raphfrk:
Ahh ok.

I think the reason is that it makes fewer changes to the craftbukkit core.

@Zeerix
(Don't think @ works to flag users, but maybe)

It looks like it is less efficient than Zeerix's original implementation.  If there are 10 players receiving the same packet, it will compress the packet 10 times.

Having said that, it probably isn't a big issue.  Packet51 packets are generally sent when a user logs in or teleports.  These inherently are for a single player.  The only time they are sent to multiple players is when a portion a chunk updates and the pull uses lower compression for those anyway.

ghost:
@Raphfrk
(@ indeed works)

The packets aren't compressed multiple times. It skips packets with valid data:
https://github.com/Bukkit/CraftBukkit/commit/70eb361891a145c776e4ec0d57596232f6483aed#L3R59

Rigby and me discussed his (first) version before he made the current one.

Raphfrk:
Ahh good point.

Anyway, the plan is to remove it from Spout once the CB RB hits.  No point in having 2 versions.

PR #49, opened 2011-07-21
*Fast* packet listeners

Nightgunner5:
I'll write JavaDocs if you like it.

This way should be faster because it's not converting `int`s to `Integer` or doing `HashMap` lookups.

### DIFF
--- a/src/org/bukkitcontrib/ContribNetServerHandler.java
+++ b/src/org/bukkitcontrib/ContribNetServerHandler.java
@@ -28,6 +28,7 @@ import org.bukkitcontrib.inventory.ContribCraftItemStack;
 import org.bukkitcontrib.inventory.ContribCraftingInventory;
 import org.bukkitcontrib.inventory.ContribInventory;
 import org.bukkitcontrib.inventory.CraftingInventory;
+import org.bukkitcontrib.packet.listener.Listeners;
 
 
 import net.minecraft.server.Container;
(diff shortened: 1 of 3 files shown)


Afforess:
Okay, I will test this when I get home from work, ~6PM EST.

Raphfrk:
Actually, does sendPacket always happen in the main thread?  If not, then the listeners array needs to be synced in some way.  Otherwise, the addListeners and sendPacket could happen in different threads.

Afforess:
> Actually, does sendPacket always happen in the main thread? If not, then the listeners array needs to be synced in some way. Otherwise, the addListeners and sendPacket could happen in different threads.

Fairly sure that it does not always happen in the main thread.

Locking with a Lock object would probably be fastest, even if it was more work.

Raphfrk:
There is an AtomicReference<V> object.

Listener would be an array of atomic references to an array of listeners.

```
final AtomicReference<Listener[]>[] listeners = new AtomicReference<Listener[]>[256];
```

The check would be

```
checkSend(int packetId) {
    Listener[] l = listeners.get()[packetId];
    <rest of code>
}
```

Since only the main thread updates the listener, the update would never actually fail.

```
listeners[packetId].set(newArray);
```

However, the more complexity, the higher the load.

Afforess:
Pulled for testing.

PR #50, opened 2011-07-22
Blank signs bug fix, doesn't compress chunks that are not in active chunk

Raphfrk:
Change list:
- (Our) Blank signs bug fix
- Doesn't compress chunks that are not active on the client
- Replaces worlds onLoad (so works if file is renamed zzzBukkitContrib.jar)
- Uses player move events to track player location (minecarts support)
- adds ReflectUtil class to handle field transfers

The minecarts things works in testing.  It generates a 5 chunk wide path for the minecart and doesn't send anything else.  I did get a client crash in the end, but I think there is something wrong with my setup.


Afforess:
Edit, Nvm

PR #51, opened 2011-07-22
Map chunk thread

Raphfrk:
Added isCancelled method call

### DIFF
--- a/src/org/bukkitcontrib/BukkitContrib.java
+++ b/src/org/bukkitcontrib/BukkitContrib.java
@@ -105,7 +105,9 @@ public class BukkitContrib extends JavaPlugin{
 		getServer().getPluginManager().registerEvent(Type.PLAYER_TELEPORT, playerListener, Priority.Monitor, this);
 		getServer().getPluginManager().registerEvent(Type.PLAYER_INTERACT, playerListener, Priority.Monitor, this);
 		getServer().getPluginManager().registerEvent(Type.PLAYER_COMMAND_PREPROCESS, playerListener, Priority.Monitor, this);
+		getServer().getPluginManager().registerEvent(Type.PLAYER_MOVE, playerListener, Priority.Monitor, this);
 		getServer().getPluginManager().registerEvent(Type.CHUNK_LOAD, chunkListener, Priority.Lowest, this);
+		getServer().getPluginManager().registerEvent(Type.WORLD_LOAD, chunkListener, Priority.Lowest, this);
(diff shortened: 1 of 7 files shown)


Raphfrk:
Probably easier to just make the change yourself :).

Btw, half day on Fridays, or was it just auto-merge?

Afforess:
Half days? I wish.

Raphfrk:
So, how did you do the merge (plus edit)?

Also, so far, it seems pretty stable with a speed of 500%, but at 1000%, the motion seems to break down.

When I am travelling, I suddenly reverse direction.  I think there is probably an integer overflow or something :).

Afforess:
> So, how did you do the merge (plus edit)?

Github has online merging tool, as well as ability to edit existing files. Handy.

> Also, so far, it seems pretty stable with a speed of 500%, but at 1000%, the motion seems to break down.

If we could find the overflow, we might be able to fix it by switching to long or double. If not... well, I generally don't recommend moving at 80 block/second anyway.

Raphfrk:
http://mc.kev009.com/Protocol#Entity_Velocity.3F_.280x1C.29

The velocity packet can't handle faster than 0.9 block per tick.  That should work out at 0.9*20 = 18 blocks per second.

Ofc, maybe it uses move packets for minecarts and the client is expected to compensate.

Afforess:
The client compensates. If you say, teleport a minecart 10 blocks when it's moving, it appears to glide through the space on the client.

It's pretty cool, actually. Set up an elevator block and watch minecarts glide through the air.

Raphfrk:
If 100% is 8 blocks per second, then 5x is 40 blocks per second, or 2 blocks per tick.  If you timed it right, you could maybe go through a brick wall.

Btw, when I click on a minecart with the client mod installed it crashes:

```
[00:53] java.lang.NullPointerException
[00:53] at net.minecraft.src.PlayerController.sendUseItem(PlayerController.java:52)
[00:53] at net.minecraft.src.PlayerControllerMP.sendUseItem(PlayerControllerMP.java:166)
[00:53] at net.minecraft.client.Minecraft.clickMouse(Minecraft.java:858)
[00:53] at net.minecraft.client.Minecraft.runTick(Minecraft.java:1035)
[00:53] at net.minecraft.client.Minecraft.run(Minecraft.java:545)
[00:53] at java.lang.Thread.run(Thread.java:662)
```

It also happens with the obfuscated client.

Raphfrk:
Anyway, had a chance to look at 500% speed?  

It is pretty clear that the 5x5 box around the player is prioritised.

Afforess:
Fixed the exception. I'm looking at your pull now.

PR #52, opened 2011-07-22
Fixes bug where signs clear on reload

Raphfrk:
On reload, the teleport smoothing mod detects a teleport.  However, the packet51 updates for teleport smoothing aren't followed by tile entity info


Raphfrk:
Hopefully last sign refresh bug :p

PR #58, opened 2011-07-24
Changes packet listeners to atomic references for thread safety

Raphfrk:
This changes the packet listeners to use atomic references.  This should give thread safety will less of the sync overhead.

Also, the way this is implemented, listeners are assigned using something like:

```
Listeners.addListener(3, this);
```

There is no packet listener manager.

Also, it might be worth sending player information for the packet.  <b>This is added in the latest version</b>

### DIFF
--- a/src/org/bukkitcontrib/ContribNetServerHandler.java
+++ b/src/org/bukkitcontrib/ContribNetServerHandler.java
@@ -19,6 +19,7 @@ import org.bukkit.Location;
 import org.bukkit.craftbukkit.entity.CraftPlayer;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Event.Result;
+import org.bukkit.entity.Player;
 import org.bukkitcontrib.event.inventory.InventoryClickEvent;
 import org.bukkitcontrib.event.inventory.InventoryCloseEvent;
 import org.bukkitcontrib.event.inventory.InventoryCraftEvent;
@@ -403,7 +404,7 @@ public class ContribNetServerHandler extends NetServerHandler{
 
 	// MapChunkThread sends packets to the method.  All packets should pass through this method before being sent to the client
 	public void sendPacket2(Packet packet) {
-		if (!Listeners.canSend(packet)) {
+		if (!Listeners.canSend((Player)player.getBukkitEntity(), packet)) {
 			return;
 		} else if(packet instanceof Packet51MapChunk) {
 			sendPacket2((Packet51MapChunk)packet);
--- a/src/org/bukkitcontrib/packet/listener/Listener.java
+++ b/src/org/bukkitcontrib/packet/listener/Listener.java
@@ -1,14 +1,16 @@
 package org.bukkitcontrib.packet.listener;
 
 import net.minecraft.server.Packet;
+import org.bukkit.entity.Player;
 
 /**
  * @author Nightgunner5
  */
 public interface Listener {
 	/**
+         * @param player The player the packet is sent to
 	 * @param packet The packet to check
 	 * @return false if the packet should be stopped, true otherwise.
 	 */
-	public boolean checkPacket(Packet packet);
+	public boolean checkPacket(Player player, Packet packet);
 }
--- a/src/org/bukkitcontrib/packet/listener/Listeners.java
+++ b/src/org/bukkitcontrib/packet/listener/Listeners.java
@@ -3,6 +3,10 @@ package org.bukkitcontrib.packet.listener;
 import java.util.Arrays;
 import net.minecraft.server.Packet;
 
+import org.bukkit.entity.Player;
+
+import java.util.concurrent.atomic.AtomicReference;
+
 /**
  * Keeps track of packet listeners
  * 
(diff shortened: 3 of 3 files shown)


Afforess:
My only fear is that this will slow packet listeners down.

Raphfrk:
I think atomic references should be reasonably fast.  However, it is definitely worth checking.

The risk with not having syncing is pretty low, since probably all the registrations will happen at startup.

I updated again to add the Player object.

Raphfrk:
Btw, how did testing go with the chunk update changes?  I was away yesterday.

desmin88:
Dislike. Arrays were used in the first place because of there fast traversing speed over lists etc.
AtomicReferences will most definitely induce a slowdown and I don't see a real need for them anyway.

Raphfrk:
> Dislike

It is an array of AtomicReferences, so you still get the array speedup.  

When Afforess tested the synchronized version, the synced version was 3x more load than the unsynced version.

Atomic references should be faster than the synced version, but slower than the non-synced version.

> I don't see a real need for them anyway

Well, they make the system thread safe.  However, if a rule is added that updates can only be made on plugin enable, then it could be safe anyway.  

Also, the pull includes the Player object associated with the packet, though, this could cause additional slowdown too.

Afforess:
> Well, they make the system thread safe. However, if a rule is added that updates can only be made on plugin enable, then it could be safe anyway. 

Thread safe, in respect to what?

There's a huge, huge difference between making listeners thread safe, as in they can call the main thread functions safely, vs. making them thread-safe to themselves, or the outside.

If I understand this code, it only makes certain that 1 listener will not be executed on 2 separate threads at the same time. Whether the listener is on the main thread or not is not observed. This provides mostly the "illusion" of safety, since calling the main thread is still unsafe.

For that reason, I see no reason to use this pull.

I will implement the player object and safety for add/removing a listener separately.

Raphfrk:
> safety for add/removing a listener separately.

That was pretty much the main point of this pull.  It was to make sure that changes to the listener maps are synced properly to the canSend method.  In canSend(), 

```
AtomicReference<Listener[]> listenerReference = (AtomicReference<Listener[]>)listeners[packet.b()];
Listener[] listenerArray = listenerReference.get();
```

this just pulls the listener array.  The AtomicReference ensures that the call either gets the old or the updated array.

In fact, it doesn't do any listener to listener syncing.  If send packet is called from 2 separate threads, then it doesn't stop the listener being called from the 2 threads.

Afforess:
Right, but it's at the expensive of performance.

I'm thinking of using a "fast-fail" system. When a listener is added or removed, it will lock the lock object. canSend will only tryLock, and if it fails, exit. That should not harm performance, while providing safety. Yes, you get the shaft if you miss a few packets, but hopefully that will only occur minimally while onEnable or onDisable.

Raphfrk:
AtomicReference tests should be pretty fast, they are "lock-free".  

If there are no writes happening, then multiple threads just can just read from the same memory location safely.

If a write does happen, it is only the reference to the array that is updated.  The read can detect if the write caused it to read corrupted data, and it just repeats the read.  If there are many more reads than writes, this reads are very fast and doesn't require the object to be locked.

I think AtomicReference.get() should be at least as fast as a call to tryLock() and won't drop any packets at all.  However, it is worth checking all methods and seeing which one is fastest in practice.

Afforess:
> If a write does happen, it is only the reference to the array that is updated. The read can detect if the write caused it to read corrupted data, and it just repeats the read. If there are many more reads than writes, this reads are very fast and doesn't require the object to be locked.

AtomicReferences are (this is an oversimplification) utility wrapper for volatile references. Simply making them volatile does not ensure safety, just that changes are reflected in new threads (instead of using a cached variable). Using volatile does not prevent the possibility of a collision between a read/write operation.

From the JDK page on Atomic package:

> "Atomic classes are designed primarily as building blocks for implementing non-blocking data structures and related infrastructure classes. The compareAndSet method is not a general replacement for locking. It applies only when critical updates for an object are confined to a single variable. "

Try lock is likely to be the approx the same speed, and will not cause the read/write failure.

Afforess:
BTW:

> The read can detect if the write caused it to read corrupted data, and it just repeats the read.

That is pretty much the definition of a lock. Locks are, as I understand them, a loop that simply checks until the lock has been freed. Exactly the same as repeating the read, really.

Raphfrk:
> Using volatile does not prevent the possibility of a collision between a read/write operation.

the .get() method on AtomicXXXX will always return a valid object, either the old or the new value.  However, for complex objects like arrays, you need to be careful.  

To update with an atomic reference, you copy the array, make your updates to the local array and then atomically update the reference so that it points to the new array, conditional on there being no change since you took the snap shot.  If it fails, you take a new snapshot and repeat.  As long as you don't update the "live" array, this is safe.

The reference is only locked for an instant at the end in order to do the swap, rather than the entire period.

An atomic reference update means that you can write at any time, but the readers can either detect corruption caused by a write, so they can repeat, or more likely the CPU can do a hardware lock for 1 clock cycle in order to do the update.  The key is that any lock is hardware rather than software and so is extremely shot.

To update an object using a read-write lock, you need to block all reads (and other writes) from happening, perform your update and then release the lock.  This stalls all read-threads for the full, read -> calculate new value -> update time period instead of just at the end for update.

Also, with read-write locks, if the object has at least 1 read-lock at all times, writes can be blocked indefinitely.  This is likely not a problem here since the server shuts down for a few ms every clock tick.

> Try lock is likely to be the approx the same speed, and will not cause the read/write failure.

Especially if you just lock the read lock, there should be very little overhead, and very few packets slipping through.  In fact, I think fully doing a read-lock rather than tryLock would be fine, since if there are no writers attempting to write, you can have an infinite number of read locks.

However, as long as you loop on failure, atomic updates will also always eventually succeed and I think my implementation will let no packets through.

Anyway, like I said, best plan is to try it a few ways and see which one is faster.

Afforess:
Just remembered I had not done this yet. Reopened to remind me.

Afforess:
Looks like it only avg's ~ 10% slower than the current design (0.04ms slower), which is fine. I'll keep it as is. :)

Raphfrk:
> Looks like it only avg's ~ 10% slower 

Cool, is interesting to see what the actual overhead is for AtomicReferences :).

PR #61, opened 2011-07-26
Made Keyboard enum more relieable on Macs.

narrowtux:
One key can now contain multiple key-codes, because some OSes tell different key-codes for keys like CMD, ALT or DELETE for example.

### DIFF
--- a/src/org/bukkitcontrib/keyboard/Keyboard.java
+++ b/src/org/bukkitcontrib/keyboard/Keyboard.java
@@ -112,7 +112,7 @@ public enum Keyboard {
 	KEY_NUMPADCOMMA(179),
 	KEY_DIVIDE(181),
 	KEY_SYSRQ(183),
-	KEY_RMENU(184),
+	KEY_RMENU(184, -72),
 	KEY_PAUSE(197),
 	KEY_HOME(199),
 	KEY_UP(200),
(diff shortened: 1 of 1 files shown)


narrowtux:
Oh, and deleted the LWIN and RWIN keys, because the were duplicated either way. When that's not okay, I can get them back...

Afforess:
Don't have a mac - but I based this purely off the LWJGL Keyboard.class. Do mac's really report those values?

narrowtux:
Yep.
I don't know if other users report the same problems, maybe I'm the only one but I don't think that.
Does it work with linux that way?

Afforess:
Since it's your pull, it's your job to test and find out if linux treats it the same. ;)

Seriously though, if you can give me 1 more mac player to back this up, I'll pull it.

K900:
The keycodes are the same on Linux. I've got an ATI GPU and I can't really test it because Minecraft runs on 5 fps with free drivers :(

Afforess:
> The keycodes are the same on Linux.

To be clear, the same as narrowtux's or the existing configuration?

K900:
Same as existing.

Afforess:
@narrowtux

First off, I don't think key codes should ever be negative. If so, it's likely LWJGL's way of saying unknown. Secondly, the negative key codes are offsets of the originals. Look at the first one you changed. KEY_RMENU. 184-256 = -72. Coincidence? Nope. All of your codes do that.

One thing you can try is grabbing the latest LWJGL binaries and Java wrappers. Notch does not use the latest versions, but they work and may fix issues with OS X.

http://lwjgl.org/download.php

K900:
That's probably some signed byte problem... again.

Afforess:
> That's probably some signed byte problem... again.

I'm not familiar with LWJGL, are signed byte issues common? (Although the real WTF is as to why Java doesn't have unsigned primitives)

I could always compensate for those numbers manually, if need be.

K900:
I ran into two or three. Though they're always the biggest pain with Java basic types :)

Afforess:
Okay, should be fixed as of this commit, @narrowtux: https://github.com/SpoutDev/Spout/commit/0f36dd2576f1ac2a35e586babbc3f93aa9898ad3

PR #66, opened 2011-07-28
PlayerInstance transfer changes

Raphfrk:
This makes minimal changes to the PlayerInstance.java and PlayerManager.java files when copying to Contrib*.java.

PlayerInstance updates in PlayerList trace the linked list of PlayerInstances.

There is a still a problem that packets in the MapChunkData queue are lost on reload (not introduced by these changes)


Raphfrk:
The best plan for actually looking at the changes would be to 

```
diff -w <craftbukkit>/PlayerInstance.java ContribPlayerInstance.java
```

 for the 2 CraftBukkit files.  That will give a better indication of the diffs compared to a diff against the previous version of the 2 files.

Raphfrk:
The latest version empties the MapChunkThread queue before reloading.

However, the chunk queue in ContribNetServerHandler is still lost.

Afforess:
> However, the chunk queue in ContribNetServerHandler is still lost.

Which is better, losing the chunk queue, or forcing the entire queue into the map thread?

Raphfrk:
> Which is better, losing the chunk queue, or forcing the entire queue into the map thread?

Well, losing the queue means that chunks won't be loaded.  Flushing the chunk queue would cause a bandwidth spike on reload.

Another option would be to move the chunk queue back to EntityPlayer.

Raphfrk:
> Another option would be to move the chunk queue back to EntityPlayer.

Should have this done in the next hour or so.

Afforess:
> Should have this done in the next hour or so.

Okay. While you're doing that, it'd be nice if you could remove the print statements

Raphfrk:
> While you're doing that, it'd be nice if you could remove the print statements

You mean in general, or just related to this pull?

Afforess:
Related to the pull.

Afforess:
Sorry about the rash of commits you will have to update for - github's online editor does not allow editing multiple files. <.<

Raphfrk:
Done.

Also, I fixed a bug in ContribCraftPlayer.  When resetting the netserver handler from Contrib back to normal, it only swapped back if the net server handlers was not the contrib version.

Afforess:
> Also, I fixed a bug in ContribCraftPlayer. When resetting the netserver handler from Contrib back to normal, it only swapped back if the net server handlers was not the contrib version.

Okay, nice find. Any chance I can get you to update this to the head commit? Makes my job easier.

Raphfrk:
Hmm, for ContribPlayerManager, was it all formatting changes?  Can I just overwrite with my version?

Afforess:
Yep - go ahead.

Raphfrk:
This doesn't compile yet, one sec :).

Raphfrk:
Seems to work now.  I did a reload while on a minecart with 500% speed and no class cast exceptions occurred.

It does cause MM to reset the speed to 100% :).

Raphfrk:
Also, worth committing these changes before changing the package names :).

PR #87, opened 2011-07-30
Changes where MapChunkThread is shutdown

Raphfrk:
Intermittent exceptions happen when reloading.

This is caused because the NetserverHandlers are switched back to normal before the MapChunkThread is shut down.

It attempts to call sendPacket2, but the method doesn't exist on normal NetServerHandlers.

This calls endThread before the netserverhandlers are switched back.  Since endThread has a .join() call, it guarantees that the MapChunkThread is shut down before the NetServerHandler replacements occur.


Raphfrk:
I never managed to replicate the PlayerInstance exception on reload from the Bukkit thread though, but they may be related.

PR #88, opened 2011-07-30
Sends key-press events for every screen in multiplayer now.

narrowtux:
Also, fixed a fix of afforess in Keyboard.getKey() when client reports negative key-id.

### DIFF
--- a/MCP/src/minecraft/net/minecraft/client/Minecraft.java
+++ b/MCP/src/minecraft/net/minecraft/client/Minecraft.java
@@ -104,6 +104,7 @@ import org.lwjgl.opengl.GL11;
 import org.lwjgl.util.glu.GLU;
 //BukkitContrib Start
 import com.pclewis.mcpatcher.mod.TextureUtils;
+import org.bukkitcontrib.gui.ScreenType;
 //BukkitContrib End
 
 public abstract class Minecraft implements Runnable {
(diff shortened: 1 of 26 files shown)


Afforess:
You can do this by just editing the GUIscreen class, which all popups extend, or intercepting them in minecraft.java instead of touching so many files

narrowtux:
Hmm, that's true.
I may fix that tomorrow, will it get into spout 0.1 then?

narrowtux:
Also, is there a way to get the player instance when you read data from a packet?
I'd like to do that to provide downwards compatibility.

Afforess:
The player Id is given in the run method. If it is done in 24 hours it will be included

narrowtux:
Is there any other way to provide backwards compatibility then? Because I send an additional int value in PacketKeyPress.

Afforess:
> Is there any other way to provide backwards compatibility then? Because I send an additional int value in PacketKeyPress.

Normally, I'd suggest you subclass PacketKeyPress and use a new, similar packet. But this is one of the few instances where breaking things is not an issue, since we are changing the namespace anyway.

halvors:
Could be possible to just take the entityid from packet and get entityid by that. Then just cast entity to player.

narrowtux:
As of changing so many files, I made a new pull request!
https://github.com/SpoutDev/Spout/pull/90
I should use the branch feature more :D

PR #102, opened 2011-08-02
Basic Packet Listener API support

Raphfrk:
There is a corresponding SpoutAPI pull request
https://github.com/SpoutDev/SpoutAPI/pull/2

This adds a support for packet listeners which don't require net.minecraft.*.

It currently only supports packets 0, 3 and 51.

New classes need to be created for all the MC standard packets.

Note: Spout shouldn't really use load: postworld, since it overrides the dependency handling.


Raphfrk:
Hmm, looks like Eclipse went crazy on import updates.

Raphfrk:
Testing involved creating a flooded world (listening for uncompressed packet 51 packets) and interfering with chat packets.

Raphfrk:
Hmm, actually I forgot the main thread resyncing stuff.

Afforess:
Okay, interesting. I'll take a look at this this weekend and write up the missing packets (fun fun fun).

Raphfrk:
> and write up the missing packets

It might be worth implementing them on a request basis.  Many probably won't be intercepted anyway.

I was also thinking of an immediateSend() method which would add the packet to the start of the queue instead of the end.  This would allow plugins to cancel packets but replace them with new ones.

Afforess:
> I was also thinking of an immediateSend() method which would add the packet to the start of the queue instead of the end. This would allow plugins to cancel packets but replace them with new ones.

Yeah, I can see the value in that.

> It might be worth implementing them on a request basis. Many probably won't be intercepted anyway.

Most of the packets shouldn't be too hard. Should probably write up one for Packet195. ;)

PR #105, opened 2011-08-03
Added resync queue and PacketManager

Raphfrk:


Afforess:
You can push this yourself. :p

Raphfrk:
Yeah, was trying to get auto-merge to work :), rather than doing a direct push that incorporates a merge.

Anyway, this adds a PacketManager class and a sendImmediatePacket method.

```
packetManager = SpoutManager.getPacketManager();

packetManager.addListener(3, new PacketListenerClass());
```

Also, it is probably worth creating a super-class and have MCPacket and SpoutPacket as a sub-class

Afforess:
> Also, it is probably worth creating a super-class and have MCPacket and SpoutPacket as a sub-class

Yeah, that's a good idea. I smell a GenericPacket. :p

Raphfrk:
I reverted the queue part.  For some reason, it was causing a large holes in the map.

Also, but less important, resyncing back to the main thread adds an average of half a clock tick of latency (and there is no guarantee that there is a heart beat every server tick).

Afforess:
Okay.

Raphfrk:
Tracked down the issue to the active chunks variable being updated to late, so new chunks are dropped.

However, adding 1 tick of latency per packet may still not be a great plan.

PR #113, opened 2011-08-04
Possible fix for NPE while polling

Top-Cat:


Raphfrk:
Hmm, am probably going to be changing that code again soon again.  It was right when I had it doing resyncing :).

tips48:
Yea, it got screwed up at build 22 i think. maybe 21, i think 22

Afforess:
Did this pull fix it?

tips48:
From what i've seen, yes. can't be sure though

PR #250, opened 2011-08-13
Double chest top/bottom detection

Majiir:
The left/right fields of double chests were arbitrarily defined by the block used to create the `DoubleChestInventory`. These commits resolve that by detecting which chests will be used to hold the top and bottom halves of the inventory.

A pull request to modify the function names in SpoutAPI is [here](https://github.com/SpoutDev/SpoutAPI/pull/5). Note that it's possible to detect which direction a double chest is facing, and thus which single chest is the left or right side, but this requires some way to determine whether a block is fully solid. Neither Bukkit nor Spout provide this, so the only way to get at it is `net.minecraft.server.Block.o[]`. If this can be resolved, I'll commit my code for detecting left/right chests and the original API functions can be used.

**EDIT 1:** I implemented the left/right detection, so the original API functions are still there. They behave differently of course, but since the prior behavior was so poorly defined, I doubt this will break many plugins.

### DIFF
--- a/src/org/getspout/spout/block/SpoutCraftChest.java
+++ b/src/org/getspout/spout/block/SpoutCraftChest.java
@@ -59,7 +59,18 @@ public class SpoutCraftChest extends CraftChest implements SpoutChest{
 	public DoubleChestInventory getFullInventory() {
 		if (isDoubleChest()){
 			SpoutCraftChest other = (SpoutCraftChest)getOtherSide();
-			return new SpoutDoubleChestInventory(new InventoryLargeChest("Double Chest", chest, other.chest), getBlock(), other.getBlock());
+			SpoutCraftChest top, bottom;
+			
+			if ((this.getLocation().getBlockX() < other.getLocation().getBlockX()) ||
+				(this.getLocation().getBlockZ() < other.getLocation().getBlockZ())) {
+				top = this;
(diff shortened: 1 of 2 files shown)


ericpauley:
Can't you just make a list of solid blocks and check blocks against it? That's what I did for an anti-xray plugin.

Top-Cat:
And what happens when a block is added, Majiir's function works, you need to release an update

Majiir:
> Can't you just make a list of solid blocks and check blocks against it? That's what I did for an anti-xray plugin.

Sure. I could also check `net.minecraft.server.Block.o[]` and be done with it. I wasn't sure if that would be considered too 'ugly' or something.

> And what happens when a block is added, Majiir's function works, you need to release an update

So should I just implement it with the NMS code and not worry about it getting into Bukkit or becoming a more general-purpose function? Do you think having the left/right functionality is necessary before pulling this?

Top-Cat:
I was more making a point, than saying whether or not that's a good idea, there is some other CB dependant stuff in here so it may not be too bad.

stephenwilliams:
Im gonna wait for @Afforess to approve this.

Afforess:
Consider it approved, I read through the code, I just cant pull it here.

PR #264, opened 2011-08-15
Biome Weather Hook

Salynne:
Added the Biome Weather hook for rendering snow/rain effects no matter the location.


Salynne:
need a rewrite, found a major bug

PR #266, opened 2011-08-15
Biome Weather Hook + Player Info framework

Salynne:
Added a BiomeManager for setting weather per player per biome or globally, to force render of specific precipitations.

Added a framework for reworking how managers keep player information.


tips48:
:D

PR #372, opened 2011-08-25
Custom texture items.

Jeckari:
This implements the server side of the custom textured items. I'm a little new to GIT, being used to SVN- I tried to keep up with the changes from the origin while I worked... hopefully that's proper procedure.

I'll also be sending along the changes to Spout-API and Spoutcraft.

As per what was mentioned, textures are individual image files.

### DIFF
--- a/src/org/getspout/spout/inventory/SimpleItemManager.java
+++ b/src/org/getspout/spout/inventory/SimpleItemManager.java
@@ -26,14 +26,17 @@ import org.bukkit.entity.Player;
 import org.getspout.spout.player.SpoutCraftPlayer;
 import org.getspout.spoutapi.inventory.ItemManager;
 import org.getspout.spoutapi.packet.PacketItemName;
+import org.getspout.spoutapi.packet.PacketItemTexture;
 import org.getspout.spoutapi.player.SpoutPlayer;
 
 public class SimpleItemManager implements ItemManager{
 	private final HashMap<ItemData, String> itemNames;
 	private final HashMap<ItemData, String> customNames;
+    private final HashMap<ItemData, String> customTextures;
 	public SimpleItemManager() {
 		itemNames = new HashMap<ItemData, String>(500);
 		customNames = new HashMap<ItemData, String>(100);
+        customTextures = new HashMap<ItemData, String>(100);
 		itemNames.put(new ItemData(1), "Stone");
 		itemNames.put(new ItemData(2), "Grass");
 		itemNames.put(new ItemData(3), "Dirt");
@@ -329,7 +332,8 @@ public class SimpleItemManager implements ItemManager{
 
 	@Override
 	public void reset() {
-		customNames.clear();
+        customNames.clear();
+        customTextures.clear();
 		for (Player player : Bukkit.getServer().getOnlinePlayers()) {
 			if (player instanceof SpoutCraftPlayer){
 				if (((SpoutPlayer)player).isSpoutCraftEnabled()) {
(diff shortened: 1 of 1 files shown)


narrowtux:
Nice. Does this have separated textures per data value?

Jeckari:
Yes, and they'll show up anywhere the item is rendered- even in achievement notifications.

narrowtux:
That's nice. I will pull it when I'm at home, when nobody other already did it :D

Top-Cat:
Looks like someone didn't get the memo about afforess hating spaces, lol (don't worry)

stephenwilliams:
@Jeckari from the Spout README.md

## Coding and Pull Request Formatting
- Generally follows the Oracle coding standards.
- Spout uses tabs, no spaces.
- No 80 column limit or midstatement newlines.
- Pull requests must compile and work.
- Pull requests must be formatted properly.
- When modifying Notch code (Minecraft vanilla code), include `//Spout start`and `//Spout end`

Please follow the above conventions if you want your pull requests accepted.

Jeckari:
Sorry about that- I just came from a project that preferred spaces over tabs, so my wires must have gotten a little crossed. I'll set my IDE to prefer tabs to spaces, and check over the source I sent in.

stephenwilliams:
@Jeckari its fine :) Just add the commits changing them to tabs when you are ready and then tag me, Afforess, Top_Cat, or Wulfspider

stephenwilliams:
oops :P

Jeckari:
Oh man, that's so cool. I make a commit and it automatically updates this pull request thing? I'm starting to like GitHub. Anyways, @alta189 I've tagged you as requested.

Jeckari:
@alta189 I've updated the Spoutcraft pull request to support standard blocks- basically anything square that uses the same texture on all six sides. Just thought you should know :)

stephenwilliams:
@Afforess so... have you looked at this?

Jeckari:
It looks like maybe @Raphfrk has implemented his own version of this kinda code? I'm not entirely certain, but a glance over recent commits seems to point in that direction- completely different implementations, but similar end results. 

I notice he calls it partial support, so maybe he's at the point I was at earlier with not getting custom textures on blocks? If that's the case, he's welcome to reference the texture-on-block code in WorldRenderer.java, and I suppose I'll get back to coding my own stuff. I do feel a bit like I've wasted time, but that's life, I guess.

stephenwilliams:
@Jeckari I am disappointed that none of the other team members ever took time to look at it and pull it or provide feedback. As I do not code the Server plugin, I did not feel that it was my place to pull it and so I didn't and asked other devs to look over it and provide feedback.

However, it seems that they never did. One of the things we pride ourselves in is being very responsive to Pull Requests, but on this one I feel that our goals were not accomplished.

Afforess:
@Jeckari

I didn't touch it because it seemed like a huge change to add to 1.0.5 which was feature locked for the time being. I was planning on looking into this more, but it seems @Raphfrk's already gotten way ahead of me and done some of it.

I do know @Raphfrk was having some issues, so perhaps he can switch over to using this instead of some of his code.

Edit:

Just to clarify - I'm in no way trying to pin blame on @Raphfrk - What happened was more of a left hand not knowing what the right hand was doing kinda thing. If anything, the failure is mine, since I'm supposed to be LEAD developer :/

Jeckari:
Oh, no worries, I don't think there's any blame to be had here. These kind of things happen in open source projects. I'm gonna write this one off as an excuse to get used to the codebase/standards and move on. As it is, it's just good to see that the feature is getting implemented. I've got a personal project that can benefit greatly from this functionality.

Anyways, if you guys happen to have other areas or features that need volunteer work, I wouldn't mind trying again. Regardless, best of luck to the project :)

Raphfrk:
Hmm, I think this is for changing textures for the current items?

The code I committed was for adding new items and allowed assigning them textures.  It isn't complete at all, it just allows swapping the texture for the item in the inventory, but it stays as Stone everywhere else.

The bulk of what I submitted yesterday was for custom item handling.  It allows you register a custom item name and get back a unique item id.  This id is used as the damage value for a Stone block.

Since the custom items are stone with various damage values, there is probably very little tweaking required to support both.

Your code can probably be used almost "out of the box" since it allows item damage values to be specified.

It might need a tweak so that it can handle "block-like" and "item-like" textures.

Jeckari:
Oh my, I'm sorry about the commotion then. I guess I jumped to conclusions... please accept my apologies. My only excuse is that it'd been a stressful couple days without power or running water.

So... where do things go from here?

Edit: Again, I feel really bad about this. @Raphfrk, my deepest personal apologies. I saw "setCustomItemTexture" and made assumptions I really should have at least tested before commenting on. If we ever meet, I owe you a few beers.

Raphfrk:
That's fine, I know what it is like to have pull requests ignored :).  Also, if you had said nothing, I may have ended up re-coding it all myself.

Changes since you have submitted them mean that they cannot be auto-merged.  I will have a look this evening, merging them manually shouldn't be that much of a problem.

Raphfrk:
Ok, I have merged the pull and hopefully, the changes I had to make were correct.

It is pretty cool, it also incorporates custom block textures.

Jeckari:
Awesome, thanks :)

Raphfrk:
Forgot to add your name to the merge message though, sorry about that.

Jeckari:
No worries, I don't really care one way or another for something as minor as that.

One thing I've been thinking about, though, is that the custom textures only allow one image file per block, which covers all sides. This works fine for 75% of blocks, and can be made to function with a bit of finagling for another 5% (doors, for example, can be made to work by overriding the texture per each data value, though it's not ideal). However, for things like pumpkins and cacti, furnaces and chests, it'd maybe be worth it to be able to set a texture to a specific face as well. I'm thinking current custom texture code sets the default texture used on all six sides, and is overridden per-face using something like the following:

setNorthFaceTexture(int blockID, short data, string url) //Sets a custom texture to only be used on a specific face

Is this worthwhile, do you think? The current functionality works in almost all cases, so it's not a big deal, but it might be nice to have that added specialization. If it is worthwhile, would you like me to look into implementing it, or should someone else work on it? It should be fairly easy to do- you'd just have to pass a different texture through to RenderBlocks.render[dir]Face()

Top-Cat:
I think it may be more worthwhile to be able to override the rending of any block on the client so that if you want it a custom shape you have control over that whereas by default it would be the usual cube.

Jeckari:
That makes more sense, yes. Good point.

Raphfrk:
With the item manager, there is a method to set a linked block (blockId and metaData).  This is the block that is placed when the custom item is used.

This system replaces the inventory texture as well as block.  It might be worth having a rule that if no linked block is set, then it appears like a sword etc (2d icon and holding).

The current texture could just be for inside the inventory and holding the item.  If it has a linked block set, the linked block should match.  Otherwise, it should appear as a 2d icon rather than a 3d block.

It might be more efficient to allow a larger texture and give the positions on the texture.  If you only had 4 variable faces, you could use a 32x32 texture.  

setNorthFaceTexture(int blockID, short data, string url, int x, int y, int h, int w) //Sets a custom texture to only be used on a specific face

x, y = position in the texture
h, w = height and width of the icon on the texture

As regards custom blocks, @Afforess is looking into those, so it is probably not worth duplicating effort.  The idea is to extend everything about the block.

PR #435, opened 2011-09-01
MCCraftPacket103SetSlot

RichardRobertson:
Implementation of MCPacket103SetSlot from this pull request: https://github.com/SpoutDev/SpoutAPI/pull/15

### DIFF
--- a/src/org/getspout/spout/packet/standard/MCCraftPacket.java
+++ b/src/org/getspout/spout/packet/standard/MCCraftPacket.java
@@ -52,6 +52,7 @@ public class MCCraftPacket implements MCPacket {
 		MCPackets[17] = MCCraftPacket17.class;
 		MCPackets[18] = MCCraftPacket18ArmAnimation.class;
 		MCPackets[51] = MCCraftPacket51MapChunk.class;
+		MCPackets[103] = MCCraftPacket103SetSlot.class;
 		
 		MCPackets[256] = MCPacket51MapChunkUncompressed.class;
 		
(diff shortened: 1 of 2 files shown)


Afforess:
I think the newline character you are using may be different. We are using whatever the windows default is.

AltyFox:
You're right.  His new line is \n (LF/Line Feed) whearas windows does \r\n  (CRFL Carriage Return Line Feed)

I personally have a STRONG distaste for the way Windows handles line breaks, and I use editors like Notepad++ that you can use Unix line breaks.

RichardRobertson:
But it doesn't change the entire file when I look at the GitHub copy of MCCraftPacket. There are a couple lines unchanged which leads me not to believe the line ending is the problem.

AltyFox:
Here's a screenshot for reference.

http://dl.dropbox.com/u/1103639/ZScreen/2011-08/Screenshot-2011-08-31_23.51.06.png

Heh.  The comments are using Unix line breaks where the code is using Windows linebreaks.  Lolwut?

RichardRobertson:
Finally fixed the line endings issue. NetBeans was converting the entire file to LF since the license header had LFs and not CRLFs. These should probably be normalized.

I had to use a different editor to preserve the line endings, but now both pull requests are ready.

Wait crap. Stupid refractor left me with an error.

RichardRobertson:
Now it's ready.

Afforess:
Great, thank you. Pulled. ;)

PR #442, opened 2011-09-01
Fix links on the readme file

GUIpsp:


lahwran:
ahhh ... this pull makes me think of the old days, when men were men and ate breakfast with a fork

PR #664, opened 2011-10-09
Implement registerItemDrop

tips48:
Not sure if this is how you want to do it, but here you go.  Should work just fine


Salynne:
There is absolutely not reason for the huge diff you have. Yes this is how I was going to do it, but I will be manually doing it.

PR #764, opened 2011-10-20
BLOCK_PLACE callEvent implementation

Hell-Fire:
This lets plugins catch custom block placements in the world, it also implements spawn protection for custom block placements

An example onBlockPlace for this is:

```
public void onBlockPlace(BlockPlaceEvent event) {
    if (event.isCancelled() || !event.canBuild()) return; // Don't do anything

    // This example shows the three levels of block types
    if(event.getBlock() instanceof SpoutBlock) {
        if (((SpoutBlock)event.getBlock()).isCustomBlock()) {
            // If you're using this, you'll probably be wanting this mostly, the "custom blocks"
            event.getPlayer().sendMessage("Placed CSB: " + ((SpoutBlock)event.getBlock()).getCustomBlock().getName());
        } else {
            // This is for blocks that Spout has extended with its own information
            event.getPlayer().sendMessage("Placed SB: " + ((SpoutBlock)event.getBlock()).getName());
        }
    } else {
        // And this is for regular old Bukkit type blocks if Spout isn't enabled
        event.getPlayer().sendMessage("Placed B: " + event.getBlockPlaced().getType());
    }
}
```

### DIFF
--- a/src/org/getspout/spout/SpoutPlayerListener.java
+++ b/src/org/getspout/spout/SpoutPlayerListener.java
@@ -158,7 +158,6 @@ public class SpoutPlayerListener extends PlayerListener{
 							canBuild = true;
 						} else {
 							Location spawn = event.getClickedBlock().getWorld().getSpawnLocation();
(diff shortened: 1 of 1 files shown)


Afforess:
Looks good.

PR #1039, opened 2011-11-28
Fix for enchantments

oloflarsson:
This fixes my bug with enchantments dissappearing when using the Inventory builder. I can not however say if the solution is the best one... Should the enchantments be referenced or cloned? I have not idea so I did clone.


narrowtux:
I think clone is more convenient.

oloflarsson:
Oh. I got pulled 0.o... ... :D/ cool.
Is there any other enchantment + inventory related issues I can help with atm?

PR #1185, opened 2011-12-09
Tool to remove ghosted custom blocks

spathizilla:
On rare occasions a customblock will fail to be removed from the world. This means all future blocks at that location will be displayed in Spoutcraft as the custom block texture. This tool removes the override when you use "/spout ghostblock" and then hit the block in question.


Top-Cat:
I feel like fixing the issue is a better idea...

spathizilla:
15:20 [Afforess(~Afforess@)] the real problem with ghost blocks is that there is no way to log all block sets
15:20 [Afforess(~Afforess@)] if we could, that bug would be gone

Was only intended until the bug could be fixed though. Easy enough to make it a standalone plugin if needed.

Afforess:
I agree it's an issue - but I want to fix it the right way. Please do release a mini-plugin in the interim though.

PR #1262, opened 2011-12-16
Fix up the pom and the start of proper tree generation routines

md-5:
The name says it all. Refactors the pom to spoutdev, automatic downloading of jterminal. The treegens become more like notch code, with a separate generator for each type. Ive just added normal for the moment. It has workaround until we know what to do with events and blockstates. All details for the treegrow event is stored in an arraylist of blockstates and instead of calling the blockbuster.update it just manually loops through and sets them all.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,16 +1,17 @@
+
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    
+
     <modelVersion>4.0.0</modelVersion>
     <groupId>net.glowstone</groupId>
     <artifactId>glowstone</artifactId>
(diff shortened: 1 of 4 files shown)


zml2008:
Would be great if you could replace all of your usages of Material with BlockID or ItemID and get/setTypeId. Lookup in Material is very slow, and speed is especially important in operations that occur frequently, such as tree generation.

md-5:
@zml2008 

lol I had that but then I thought for changes to block id etc it would be better

md-5:
done

zml2008:
Well we're trying to have no dependencies on Bukkit as well (SpoutAPI is
going to replace Bukkit as a plugin API)

md-5:
I can try and remove more, but at the moment they are only there because they are what Glowstone is passing around

zml2008:
If Glowstone's passing them around, feel free to replace them with ID's

md-5:
Well we're trying to have no dependencies on Bukkit as well (SpoutAPI is going to replace Bukkit as a plugin API)

Glowstone is passing round Bukkit types in a lot of places

md-5:
redoing

PR #1271, opened 2011-12-16
Added /give command!

ZNickq:
Works a bit like /i in some bukkit plugins:
/give znickq stone 10 -> gives znickq 10 stone!
/give stone 10 -> gives the player 10 stone!
/give stone -> gives the player 64 stone!

Only works for ops right now! :)

### DIFF
--- a/src/main/java/org/getspout/server/SpoutServer.java
+++ b/src/main/java/org/getspout/server/SpoutServer.java
@@ -445,7 +445,7 @@ public final class SpoutServer implements Server {
 		commandMap.register(new SayCommand(this));
 		commandMap.removeAllOfType(ReloadCommand.class);
 		commandMap.register(new ReloadCommand(this));
-        commandMap.register(new GiveCommand(this));
+                commandMap.register(new GiveCommand(this));
 		commandMap.register(new HelpCommand(this, commandMap.getKnownCommands(false)));
 
 		enablePlugins(PluginLoadOrder.STARTUP);
--- a/src/main/java/org/getspout/server/SpoutServer.java
+++ b/src/main/java/org/getspout/server/SpoutServer.java
@@ -55,22 +55,7 @@ import org.bukkit.plugin.SimpleServicesManager;
 import org.bukkit.plugin.java.JavaPluginLoader;
 import org.bukkit.util.permissions.DefaultPermissions;
 
-import org.getspout.server.command.BanCommand;
-import org.getspout.server.command.ColorCommand;
-import org.getspout.server.command.DeopCommand;
-import org.getspout.server.command.GameModeCommand;
-import org.getspout.server.command.SpoutCommandMap;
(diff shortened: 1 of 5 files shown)


src/main/java/org/getspout/server/SpoutServer.java
```diff
@@ -55,22 +55,7 @@
...
-import org.getspout.server.command.ListCommand;
-import org.getspout.server.command.MeCommand;
-import org.getspout.server.command.OpCommand;
-import org.getspout.server.command.ReloadCommand;
-import org.getspout.server.command.SaveCommand;
-import org.getspout.server.command.SayCommand;
-import org.getspout.server.command.StopCommand;
-import org.getspout.server.command.TimeCommand;
-import org.getspout.server.command.WhitelistCommand;
+import org.getspout.server.command.*;
```
Top-Cat: Nope, nope, nope

ZNickq: Fixing it now ;)
Netbeans does it automatically :/ Need to find a way to turn it off...

macalinao: Doesn't do it for me.

src/main/java/org/getspout/server/SpoutServer.java
```diff
@@ -460,6 +445,7 @@ public void start() {
 		commandMap.register(new SayCommand(this));
 		commandMap.removeAllOfType(ReloadCommand.class);
 		commandMap.register(new ReloadCommand(this));
+                commandMap.register(new GiveCommand(this));
```
Afforess: Tabs vs. Spaces.

ZNickq: Yeah, i fixed it in the next commit :)
Will work on fixing any other problems now!

src/main/java/org/getspout/server/SpoutServer.java
ZNickq: o_O when i add a certain number of imports, it just switches to *...wierd, i'll google it ;)

ZNickq:
Ok, this should be it! I set netbeans to no longer do wildcards too! :)

Afforess:
I think you should write this command so it can be executed from the console too.

ZNickq:
@Afforess Ok, did that, also added /tp! I think my formatting is better now! :)

Top-Cat:
Also you need a merge

RoyAwesome:
Since we restructured Spout, this is irrelevant now.  Thanks though.

ZNickq:
Np, but how did you restructure it? o-o

RoyAwesome:
Blocks are going into Vanilla, so /give should probably live there (or take any registered block).  Not sure how to do that right now since blocks aren't in yet, but, we are getting there.

PR #1273, opened 2011-12-16
Added Workbench interaction handling

lopcode:
Right clicking workbench works as expected, window pops up, window handling on the backend is done as expected.

Workbench crafting still not implemented - someone smarter than me should take a look at WindowClickMessageHandler to call craft() appropriately.

CraftingInventory.craft() is not inventory agnostic - needs looking at so it works for different types of inventory.


ZNickq:
Nice! :)

narrowtux:
Is your copy on a FAT-Volume? I'd strongly advise against that, because you changed all the file modifiers to 100775. This will cause strange effects on the console because it thinks that they are executeable.

lopcode:
I'm on Windows, NTFS, with NetBeans. I have absolutely no idea why the file modes are changing :S

zml2008:
This would create a StackOverflowException any time someone placed a workbench. Did you even test this?

lopcode:
... it works fine for me? EDIT: Confirmed, there's no exceptions.

narrowtux:
It generates this exception when I close the Inventory screen:
http://pastebin.com/m2SYNy4y

lopcode:
Someone else was looking at fixing the close window bug in the channel - I didn't touch it.

ZNickq:
It was me ;) 
Problem is,  i am not sure where in spout the opening of windows takes place...

zml2008:
I removed some things, and the inventory + block physics APIs are probably going to be refactored a ton. Also, your formatting is wrong. You can look at every other source file in the project for correct examples. For now I'll be closing this.

PR #1276, opened 2011-12-16
Added /togglestorm command

aPunch:
This adds the vanilla /toggledownfall command to Spout, but with a better name.


zml2008:
Merged in e00cb838362b5f95280309eec76181abadec5709

PR #1281, opened 2011-12-17
Command Additions

aPunch:
This pull contains:
-/xp command
-/tell command
-Misc. cleanup


ZNickq:
Nice!

narrowtux:
@aPunch in the tell command, can you make sure that the sender also sent a message? It would be a bit bad if they can send empty messages :P
Other than that, it looks good.

aPunch:
Fixed!

PR #1290, opened 2011-12-18
Mob Stuff

aPunch:
In this pull request:

-Updated mob drops to latest vanilla spec
-Removed unused imports in entity classes
-Standardized constructor javadocs in entity classes


tips48:
Nice job!

PR #1305, opened 2011-12-20
Messing with formatting

Top-Cat:
Made it a PR to get some feedback

**DO NOT PULL**


Top-Cat:
I would mention I haven't looked at every file, git makes it nicer to look through too

PR #1303, opened 2011-12-19
Added /teleport and /kill command

RafVDL:
Added vanilla teleport (alias: tp, tele) command to Spout.
/tp playerfrom playerto
or /tp playerto (if you run it from a client).

Both can be ran in the client, but only the first one in console.

Added vanilla kill command to Spout.
/kill

Only client, only OP.

The SpoutServer.java commit got screwed up :P

/me going to bed now :D

PR #1309, opened 2011-12-20
Crafting Recipes and /kill Command

aPunch:
In this pull request:

-/kill command using suggestions from @NinjaZidane in this pull: https://github.com/SpoutDev/Spout/pull/1303
 (entity.damage() is not implemented in Spout, but once it is this should work)
-Added many crafting recipes (there are probably still some missing, but I added as many as I could using the Minecraft wiki)


Zidane:
Cheers aPunch :)

RoyAwesome:
Since we restructured Spout, this is irrelevant now.  Thanks though!

PR #1312, opened 2011-12-20
Added /teleport command. Allow custom kick message in stop command.

RafVDL:
This time without a screwed up commit :D

Added /teleport (alias: tp) command:
- Usage:
  /tp <playerfrom> <playerto>
  /tp <playerto> // Only from client.

Wrong number of arguments => blocked.

/shutdown [message]
All information in commit message :P
Tested (on commit 13e2c3ceff4a95b9b1ed163424670a27b5c5d512) and works, ofcourse.

### DIFF
--- a/src/main/java/org/getspout/server/SpoutServer.java
+++ b/src/main/java/org/getspout/server/SpoutServer.java
@@ -499,9 +499,11 @@ public final class SpoutServer implements Server {
 	}
 
 	/**
-	 * Stops this server.
+	 * Stops this server with a given kick message.
+	 * 
+	 * @param message The message to display as kick message.
 	 */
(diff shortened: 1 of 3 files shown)


lukespragg:
Did not mean to close that. Wrong button. :/

RafVDL:
Added /stop [message]
All information in commit message (ca31488) :P

windy1:
Only concern is I'm unaware of how close Spout wants to stay to the vanila, otherwise, this could be a good idea, but also could be done as a plugin.

aPunch:
It seems this is close enough to vanilla for Spout's tastes. I know some other commands are "less vanilla" than this.

PR #1318, opened 2011-12-21
Expensive lookup method

Nican:
SpoutWorld.getHighestBlockYAt would lookup for the same chunk over and over again, now it is just one lookup, and checks the whole height.


Raphfrk:
You raise another point.  We are moving to cubic Chunks ... so I think this needs to be re-considered.

Nican:
I just started taking a look at the minecraft and the code is still a bit confusing for me, but why not instead of having SpoutWorld save every single loaded chunk, have the io application track the blocks that was changed, and just write to file those parts?
Java.nio provides some really nifty tools for locking and writing specific parts of a file.

RoyAwesome:
We are moving to Cubic Chunks, so this is not compatible.  Thanks though!

PR #1389, opened 2012-01-05
Implement Permission methods in Player and CommandSource

yetanotherx:
This is a co-PR with https://github.com/SpoutDev/SpoutAPI/pull/59. Per feedback in IRC, I've implemented an extensible permissions system. This is the actual implementation of the interfaces that are described in the API pull request. 

Questions? #spoutdev as usual.


zml2008:
Added in 7d56cd5cbfd9c93ff6bab0814d811b2823b2bc0a

PR #1423, opened 2012-01-12
Implemented basic inventory handling...

Doridian:
...and currently held item for inventories.

### DIFF
--- a/src/main/java/org/spout/server/SpoutBlock.java
+++ b/src/main/java/org/spout/server/SpoutBlock.java
@@ -1,0 +1,59 @@
+package org.spout.server;
+
+import org.spout.api.geo.World;
+import org.spout.api.geo.cuboid.Block;
+import org.spout.api.material.BlockMaterial;
+import org.spout.api.material.MaterialData;
+
+public class SpoutBlock extends Block {
+	BlockMaterial material = null;
+
+	public SpoutBlock(World world, int x, int y, int z) {
+		super(world, x, y, z);
+	}
+	
+	public SpoutBlock(World world, int x, int y, int z, short id) {
+		this(world, x, y, z);
+		setBlockId(id);
+	}
+
+	public SpoutBlock(World world, int x, int y, int z, BlockMaterial material) {
+		this(world, x, y, z);
+		setBlockMaterial(material);
+	}
+	
+	@Override
+	public BlockMaterial setBlockMaterial(BlockMaterial material) {
+		setBlockId(material.getId());
+		return material;
+	}
+
+	@Override
+	public short setBlockId(short id) {
+		BlockMaterial newMat = MaterialData.getBlock(id);
+		if(newMat != null) {
+			material = newMat;
+		}
+		return material != null ? material.getId() : 0;
+	}
+
+	@Override
+	public BlockMaterial getBlockMaterial() {
+		return material;
+	}
+
+	@Override
+	public short getBlockId() {
+		return material.getId();
+	}
+
+	@Override
+	public BlockMaterial getLiveBlockMaterial() {
+		return getBlockMaterial();
+	}
+
+	@Override
+	public short getLiveBlockId() {
+		return getBlockId();
+	}
+}
--- a/src/main/java/org/spout/server/entity/SpoutEntity.java
+++ b/src/main/java/org/spout/server/entity/SpoutEntity.java
@@ -25,7 +25,7 @@
  */
 package org.spout.server.entity;
 
-import java.io.Externalizable;
+import java.io.Serializable;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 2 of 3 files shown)


RoyAwesome:
can you update the pull to these guidelines? http://forums.spout.org/threads/license-and-pull-requests.643/ 

(Notably 1 commit and the signoff message)

src/main/java/org/spout/server/SpoutBlock.java
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	public SpoutBlock(World world, int x, int y, int z, BlockMaterial material) {
+		this(world, x, y, z);
+		setBlockMaterial(material);
+	}
+	
+	@Override
+	public BlockMaterial setBlockMaterial(BlockMaterial material) {
+		setBlockId(material.getId());
```
Raphfrk, line 27: This should probably also set the block's data and auxData.

We haven't implemented that yet though.

I am planing to add a BlockFullState class that will have the 3 pieces of info.

src/main/java/org/spout/server/SpoutBlock.java
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	@Override
+	public short setBlockId(short id) {
+		BlockMaterial newMat = MaterialData.getBlock(id);
+		if(newMat != null) {
+			material = newMat;
+		}
+		return material.getId();
+	}
```
Raphfrk: Could be a NPE here, if material is null?

src/main/java/org/spout/server/SpoutChunk.java
```diff
@@ -342,6 +342,9 @@ public void populate(boolean force) {
 			}
 		}
 		populated.set(true);
+		if(getRegion() instanceof SpoutRegion) {
+			((SpoutRegion) getRegion()).onChunkPopulated(this);
+		}
```
Raphfrk: I think @narrowtux  already did chunk population?

Raphfrk:
I would also like it updated to the guidelines too.

src/main/java/org/spout/server/SpoutRegion.java
```diff
@@ -78,6 +82,15 @@
 	 */
 	protected final EntityManager entityManager = new EntityManager();
 	
+	/**
+	 * Holds all not populated chunks
+	 */
+	protected Set<Chunk> nonPopulatedChunks = new HashSet<Chunk>();
```
Raphfrk: This should probably be a ConcurrentHashMap so as to be thread safe

src/main/java/org/spout/server/entity/SpoutEntity.java
```diff
@@ -270,6 +271,26 @@ public void setData(String key, Externalizable value) {
...
+	}
+
+	private Inventory inventory;
+	@Override
+	public Inventory getInventory() {
+		if(inventory == null) {
+			SpoutDatatableObject obj = (SpoutDatatableObject)getData("inventory");
+			if(obj == null) {
+				inventory = new Inventory(getInventorySize());
+				setData("inventory", inventory);
```
Raphfrk, line 304: Hmm, so all entities would have an inventory?  Maybe there should be a (boolean create) parameter to indicate if one should be created.

Doridian, line 304: Could also just override the getInventorySize() method so it returns 0 (and then implement if into getInventory() that returns null in such cases?)

src/main/java/org/spout/server/SpoutBlock.java
Raphfrk, line 27: I have added the class to SpoutAPI, so you can get all 3 pieces of info at once.  You will also be able to set all 3 at once.

src/main/java/org/spout/server/entity/SpoutEntity.java
Raphfrk, line 304: Another thought is that your method causes a string lookup every time.

DataTables allow an id to be used as an alternative.  

Something like:

```
static volatile inventoryId;
static volatile inventoryIdSet = false;
...
...
...
if (!inventoryIdSet) {
    inventoryId = stringMap.register("inventory");
    inventoryIdSet = true;
}
```

There would need to be a way to get the DataTable string map though.  @RoyAwesome may have set up how to do it.

RoyAwesome, line 304: Yeah, I'll do that now, it's a good idea.

Doridian, line 304: Also my method would only cause a lookup once per entity (when it first stores it into the "local" variable inventory)

Raphfrk, line 304: True, I was just aiming to get the principle of using the ids out there :)

PR #1492, opened 2012-01-24
Fix for random ConccurentModificationException when walking around!

ZNickq:
I can't test it because i don't have any compiler, but it should work!
This is the stack trace which shows the error: http://pastie.org/3241897


ZNickq:
Forgot to sign it off :P Easier just making another one ;)

PR #1493, opened 2012-01-24
Fix for random ConccurentModificationException when walking around!

ZNickq:
I do not have a compiler to test it, but it should work!
This is the stack trace: http://pastie.org/3241897


ZNickq:
Oh, and i think i should tag @Raphfrk for this! ^^

Raphfrk:
It was @narrowtux who did populating.  I think a concurrent hashmap is right though.

A chunk can be loaded from anywhere, so needs to be able to be added to the queue from 2 different threads.

Populating itself should happen in a region thread specific way.  However, some support for cross region effects is required.

PR #1517, opened 2012-01-28
Falling sand!

Darsstar:
No animation though =[


Afforess:
Not the exact correct solution, but thanks.

PR #1570, opened 2012-02-07
Better command registration

windy1:
A bit more developer friendly - Respective SpoutAPI PR: https://github.com/SpoutDev/SpoutAPI/pull/81

Signed-off-by: WalkerCrouse wjcrouse913@aim.com

### DIFF
--- a/src/main/java/org/spout/server/SpoutServer.java
+++ b/src/main/java/org/spout/server/SpoutServer.java
@@ -233,6 +233,8 @@ public class SpoutServer extends AsyncManager implements Server {
 	 * Cached copy of the server configuration, can be used instead of re-parsing the config file for each access
 	 */
 	private Configuration configCache = null;
+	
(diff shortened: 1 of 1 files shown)


zml2008:
This doesn't work. Discussion in IRC:

```
<zml2008> That avoids one of the major points of the commands system...
<Windwaker> how so?
<Windwaker> maybe im missing the point
<zml2008> The point is that using the annotated command registration system is only one way to register commands, and the way you implemented it requires command methods to have specific constructor arguments too
<zml2008> That will also force plugin developers to have static references to their plugins
<Windwaker> Yes I'm aware there will be CommandExecutors and what not, but this just provides access to easily add sub commands
<zml2008> Node the new SimpleInjector(this) in SpoutServer
<zml2008> with that, all command classes registered using your new method would have to have a single argument object constructor that takes a SpoutServer
```

PR #1571, opened 2012-02-07
Service manager implemented

Sleaker:
Implements the Service Manager for use in SpoutServer


Salynne:
This just needs to be signed off and awaits the other pull.

Sleaker:
fixing.

PR #1617, opened 2012-02-18
Fixed the Session

steffengy:
The session now deletes a player in all minecraft's of the users , which could be able to see him.


Top-Cat:
Multiple unsigned commits, please combine and sign.
You can sign with the git command line with -s and this shows you agree to our code license.

Salynne:
Thanks :)

PR #1623, opened 2012-02-18
Fixed  a Y problem, while falling under bedrock

steffengy:
Fix of some produced exceptions, while falling down to like -700, where the server tried to create this chunk.


ZNickq:
-___-
It's supposed to populate chunks above and below bedrock :/

Afforess:
@ZNickq is correct. I just have not had the free time to implement infinite height yet for clients.

PR #1629, opened 2012-02-19
Ban feature

greatman:
Added ban system using the spout.yml file.

### DIFF
--- a/src/main/java/org/spout/server/SpoutServer.java
+++ b/src/main/java/org/spout/server/SpoutServer.java
@@ -682,6 +682,7 @@ public class SpoutServer extends AsyncManager implements Server {
 		for (int i = 0; i < whitelist.length; i++) {
 			whitelist[i] = whitelistedPlayers.get(i);
 		}
+		
 		return whitelist;
 	}
 
@@ -787,17 +788,16 @@ public class SpoutServer extends AsyncManager implements Server {
 	}
 
 	@Override
-	public void ban(String address) {
-		// TODO Auto-generated method stub
-
+	public void ban(Player player) {
+		bannedPlayers.add(player.getName());
 	}
 
 	@Override
-	public void unban(String address) {
-		// TODO Auto-generated method stub
-
+	public void unban(Player player) {
+		bannedPlayers.remove(player.getName());
 	}
 
+	
 	@Override
 	public List<String> getAllPlayers() {
 		// TODO Auto-generated method stub
@@ -844,10 +844,10 @@ public class SpoutServer extends AsyncManager implements Server {
 	}
 
 	@Override
-	public Collection<Player> getBannedPlayers() {
-		Set<Player> players = new HashSet<Player>();
+	public Collection<String> getBannedPlayers() {
+		Set<String> players = new HashSet<String>();
 		for (String name : bannedPlayers) {
-			players.add(new SpoutPlayer(name));
+			players.add(name);
 		}
 		return players;
 	}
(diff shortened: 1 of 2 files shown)


src/main/java/org/spout/server/SpoutServer.java
```diff
@@ -787,17 +788,66 @@ public World getDefaultWorld() {
...
-		// TODO Auto-generated method stub
-
+	public void ban(Player player) {
+		
+		bannedPlayers.add(player.getName());
+		Configuration config;
+		
+		try
+		{
+			config = getConfiguration(true);
```
Afforess: Is there a reason you need to FORCE a configuration update with that true parameter?

src/main/java/org/spout/server/SpoutServer.java
```diff
@@ -787,17 +788,66 @@ public World getDefaultWorld() {
...
+			
+			if (bannedList == null) {
+				bannedList = new ArrayList<String>();
+			}
+			
+			if (!bannedList.contains(player.getName())) {
+				bannedList.add(player.getName());
+			}
+				
+			config.setProperty("banlist", bannedList);
```
Afforess: You don't need to set the property unless it was null.

src/main/java/org/spout/server/SpoutServer.java
```diff
@@ -787,17 +788,66 @@ public World getDefaultWorld() {
...
+			config.save();
```
Afforess: Is there a reason we need to save this right away? Wouldn't saving at shutdown be acceptable? File IO is terribly slow.

src/main/java/org/spout/server/SpoutServer.java
```diff
@@ -787,17 +788,66 @@ public World getDefaultWorld() {
...
+	public void unban(Player player) {
 		// TODO Auto-generated method stub
-
+		String playerName = player.getName();
+		bannedPlayers.remove(playerName);
+		Configuration config;
+		
+		try
+		{
+			config = getConfiguration(true);
```
Afforess: Same comments as above.

src/main/java/org/spout/server/SpoutServer.java
greatman: So basicly just add into the bannedList variable and the server takes care of the rest?

Afforess: Right. The configuration reference will be up to date still. You don't need to recheck the file every time.

Salynne:
Talked about this a bit on IRC and came to some decisions.

1) I don't agree that it should wait til shutdown to save. Afforess may come back and override me on that so don't worry about that right now. The reasoning is that you do not save very often and you do not want to crash and lose bans. Banned people coming back is typically bad.

2) I would like to see IP bans still being possible alongside player bans, this prevents stuff like some crappy DDOS-Style stuff I have been hit with before, among other things.

3) The ban/whitelist/iplist should not be in the main spout.yml. They should be in their own individual yaml or txt file.

I at least want 3 to exist to pull, if you can do 2 for me that'd be great, and we'll have to wait to see on 1 because I don't want to reverse his decision and have you change it and then have him reverse it again. THat'd be silly :P

Afforess:
> 1) I don't agree that it should wait til shutdown to save. Afforess may come back and override me on that so don't worry about that right now. The reasoning is that you do not save very often and you do not want to crash and lose bans. Banned people coming back is typically bad.

Yeah I wanted 300 second saving intervals.

greatman:
I can work out 2 and 3. Ill let you guys work on 1 because I know I will
have a bad implementation lol.

Salynne:
Closing these, open a new one if you've updated them.

PR #1663, opened 2012-02-27
Initial MaterialEntity implementation!

ZNickq:
SpoutAPI PR:  https://github.com/SpoutDev/SpoutAPI/pull/95
This is pretty much what was discussed in the irc log, gotta tell you, love it more than my old PR :D

### DIFF
--- a/src/main/java/org/spout/server/SpoutBlockm.java
+++ b/src/main/java/org/spout/server/SpoutBlockm.java
@@ -25,6 +25,7 @@
  */
 package org.spout.server;
 
+import org.spout.api.entity.MaterialController;
 import org.spout.api.geo.World;
 import org.spout.api.geo.cuboid.Blockm;
 import org.spout.api.material.BlockMaterial;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spout/server/SpoutBlockm.java
```diff
@@ -76,4 +77,14 @@ public BlockMaterial getLiveBlockMaterial() {
 	public short getLiveBlockId() {
 		throw new UnsupportedOperationException("Operation is not supported");
 	}
+
+	@Override
+	public void setMaterialEntity(MaterialController materialEntity) {
+		throw new UnsupportedOperationException("Operation is not supported");
```
Afforess, line 83: These should actually all get implemented. The reason they weren't was because the original submitter was lazy - not because of design.

ZNickq, line 83: Well, i'm not sure what SpoutBlockm actually does to be honest :P I would gladly implement it otherwise :)

ZNickq:
Closing this since it will be done another way ;)

PR #1667, opened 2012-02-28
Implemented getClosestPlayer!

ZNickq:
SpoutAPI PR: https://github.com/SpoutDev/SpoutAPI/pull/97

### DIFF
--- a/src/main/java/org/spout/server/SpoutWorld.java
+++ b/src/main/java/org/spout/server/SpoutWorld.java
@@ -67,52 +67,42 @@ import org.spout.server.util.thread.snapshotable.SnapshotableReference;
 public class SpoutWorld extends AsyncManager implements World {
 
 	private SnapshotManager snapshotManager = new SnapshotManager();
-
 	/**
 	 * The server of this world.
 	 */
 	private final Server server;
-
 	/**
 	 * The name of this world.
 	 */
 	private final String name;
-
 	/**
 	 * The region source
 	 */
 	private final RegionSource regions;
-
 	/**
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/server/SpoutWorld.java
```diff
@@ -509,13 +499,26 @@ public void removePlayer(Player player) {
...
 	}
-	
-	
+
+	@Override
+	public Player getClosestPlayer(Point p) {
+		Player result = null;
+		float minDistance = -1;
+		for (Player plr : players) {
+			float dist = p.subtract(plr.getEntity().getPoint()).length();
```
narrowtux, line 513: length is using sqrt and thus is a lengthy operation.
Seeing that you want to call this for each dropped item each tick, this will have a negative effect on speed.
You might consider using .lenghtSquared() for that.

ZNickq, line 513: http://pastie.org/3492055 
Is this faster? It's using a different method :P

Darsstar, line 513: Nope, still uses Math.sqrt()

ZNickq, line 513: Well, there's no distanceSquared yet :/

Darsstar, line 513: Point.getSquaredDistance(Point other)

Salynne:
This needs to be updated to use the getSquaredDistance as well as updated to be merged again.

PR #1668, opened 2012-02-28
Basic changes to SpoutEntity!

ZNickq:
Fixed local variable names
Added missing @Overrides
Allow settings of a null Controller
Remove entity from EntityManager after it is killed
kill() now properly returns true if the entity was successfully killed.


ZNickq:
I misunderstood what kill() returned :P All the other changes were good though, will make a new PR for em
LE: Actually, i'll wait for Olloth to come back, need to make sure i get this right :)

PR #1672, opened 2012-02-28
Added some checks to prevent error spam!

ZNickq:
:)
Apparently this shouldn't be needed, so the error is probably somewhere else, but it can't hurt :P

### DIFF
--- a/src/main/java/org/spout/server/entity/SpoutEntity.java
+++ b/src/main/java/org/spout/server/entity/SpoutEntity.java
@@ -300,8 +300,11 @@ public class SpoutEntity implements Entity {
 				((SpoutChunk) chunk).removeEntity(this);
 			}
 			if (chunkLive == null) {
-				((SpoutChunk) chunk).removeEntity(this);
-				entityManagerLive.deallocate(this);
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/server/entity/SpoutEntity.java
```diff
@@ -300,8 +300,11 @@ public void finalizeRun() {
 				((SpoutChunk) chunk).removeEntity(this);
 			}
 			if (chunkLive == null) {
-				((SpoutChunk) chunk).removeEntity(this);
-				entityManagerLive.deallocate(this);
+				if(chunk!=null&&chunk.isLoaded())
+					((SpoutChunk) chunk).removeEntity(this);
```
Darsstar, line 304: I changed my mind about that `chunk` can cause an NPE.

That code is inside a `if (chunkLive != chunk) {`block. so if `chunkLive` is null, `chunk`can't be.

Darsstar, line 304: Actually, line 300 already took care of it (line 304). So those two lines can be removed.

Afforess:
Let me know if any of these changes are still needed.

PR #1674, opened 2012-02-29
Allow people to actually bind ip/port

katuma:
(unless theres plugin even for that which i might've missed?)

### DIFF
--- a/src/main/java/org/spout/server/entity/SpoutEntity.java
+++ b/src/main/java/org/spout/server/entity/SpoutEntity.java
@@ -28,7 +28,6 @@ package org.spout.server.entity;
 import java.io.Serializable;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
-
 import org.spout.api.collision.CollisionModel;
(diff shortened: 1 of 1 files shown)


zml2008:
Looks like this was missed when migrating to the updated configuration. Will test and pull.

zml2008:
Hm, this doesn't seem to work. I changed the bind port with this fix and the server still attempted to use the default. Did this work for you?

katuma:
sure, otherwise i'd not be sending a pull :)

you have to put something like this

address: 0.0.0.0:20000

in your config/config.yml

zml2008:
ah, looks like somebody changed the config file while I wasn't looking. Your commit seems to fix the issue, but you'll have to commit it with the Signed-Off-By text (you can get this with git commit -s). More information about that is at http://forums.spout.org/threads/license-and-pull-requests.643/

DrakiaXYZ:
But @zml2008 I need this so I can start the server (Alongside my main server), and I don't want to compile it myself! :P

katuma:
lol zml2008, who needs signoffs ... cut&paste works as well for bukkit as spout! j/k, new pull against head

PR #1681, opened 2012-03-01
Implemented /give.

spacechase0:
It crashes the client whenever I try to give something that's a GenericTool or GenericWeapon from Vanilla. Not sure why, but sometimes it tells me it's a NullPointerException.

Signed-off-by: Chase W chasew@silkweaver.com

### DIFF
--- a/src/main/java/org/spout/server/command/AdministrationCommands.java
+++ b/src/main/java/org/spout/server/command/AdministrationCommands.java
@@ -36,6 +36,10 @@ import org.spout.api.command.CommandSource;
 import org.spout.api.command.annotated.Command;
 import org.spout.api.command.annotated.CommandPermissions;
 import org.spout.api.exception.CommandException;
+import org.spout.api.inventory.Inventory;
+import org.spout.api.inventory.ItemStack;
+import org.spout.api.material.Material;
+import org.spout.api.material.MaterialData;
 import org.spout.api.player.Player;
 import org.spout.server.SpoutServer;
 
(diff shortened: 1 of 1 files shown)


Zidane:
Looks pretty good...I want to take this farther though and allow strings to be used as well.

Idea is you could do /give NinjaZidane wool 1 which would give the first registered material named "Wool" to the player...but what if two plugins have the same material named wool? /give NinjaZidane wool:1 1 would suffice I think. Only issue I think is when you have a block that has different states (aka redwool)...personally I think we can just create a new block for them and you can do /give NinjaZidane redwool 1 then.

Any ideas from anyone else?

spacechase0:
Strings actually do work a little ([see here](https://github.com/spacechase0/Spout/commit/a57e2dee4ca9e0abc0ce1a082f2c476bcb7044dc#L0R110)), but any block with a space in it (like Red Mushroom) won't work, since arguments are separated by spaces.

It looks [from here](https://github.com/SpoutDev/SpoutAPI/blob/master/src/main/java/org/spout/api/material/GenericItemMaterial.java#L43) that the name and display name use the same string. Maybe replacing spaces in name (but not display name) would fix that.

Maybe another way for specifying which plugin would be prefixing the plugin name, like vanilla:wool?

SonnyX:
A way to make items with spaces work is check the amount of Integers and use the latest for the amount and process the first 2 the same as you normally would.

windy1:
Ehm... this should be in Vanilla probably.

spacechase0:
How come? I don't understand why. :P

It would probably be useful for games that don't use Vanilla at all, too.

Salynne:
I agree the command implement should go in vanilla, but the string lookup should be a feature of spout. Also I like your suggestion, we should space spaces with _ or something in the registry for easy command usage. Totally forgot about that.

Salynne:
Closing this because I mostly forgot about it. The whole material system is being revamped currently and then following that the concept of a give command will have to be revisited. Sorry.

PR #1689, opened 2012-03-02
Added SpoutWorld.getEntity(int id)

Darsstar:
See https://github.com/SpoutDev/SpoutAPI/pull/104 for API

### DIFF
--- a/src/main/java/org/spout/server/SpoutWorld.java
+++ b/src/main/java/org/spout/server/SpoutWorld.java
@@ -262,9 +262,7 @@ public class SpoutWorld extends AsyncManager implements World {
 		if (e.isSpawned()) {
 			throw new IllegalArgumentException("Cannot spawn an entity that is already spawned!");
 		}
-		SpoutRegion region = (SpoutRegion) e.getRegion();
-		region.addEntity(e);
-
+		((SpoutRegion) e.getRegion()).addEntity(e);
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/server/SpoutWorld.java
```diff
@@ -118,6 +114,16 @@
 	 */
 	private final Set<Player> players = Collections.newSetFromMap(new ConcurrentHashMap<Player, Boolean>());
 
+	/**
+	 * A queue of all entities to be acclocated in the entity manager
```
ZNickq: acclocated? o.O

src/main/java/org/spout/server/SpoutWorld.java
```diff
@@ -25,13 +25,9 @@
  */
 package org.spout.server;
 
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
Darsstar: bah, ty sonny

src/main/java/org/spout/server/SpoutWorld.java
SonnyX: accLocated:
A method that retrieves the specified object's current screen location.

src/main/java/org/spout/server/SpoutWorld.java
SonnyX: np ^^

src/main/java/org/spout/server/SpoutWorld.java
ZNickq: I am pretty sure it's supposed to be allocated though, lawl (since it's the spawn Queue ;P)

PR #1708, opened 2012-03-08
Added a NetBeans license hint to the pom

Darsstar:
see http://wiki.netbeans.org/MavenBestPractices#License_Header

Is this this @Wulfspider approved?


lukespragg:
This isn't really necessary, being as we may be pushing project files for each. This entry in the pom would only apply to anyone who actually manually made that license in their IDE, so a full project file would be better.

Darsstar:
Oh, that works too =]
And Spout already has one it seems...

PR #1732, opened 2012-03-11
Fixed the SpoutServer to start the StorageQueue.

steffengy:
Fixes throwing of exceptions at the start of the server.

StorageQueue always throws: "Cannot queue tasks while thread is not running"
because it was never started.

Signed-off-by: steffen steffenb198@aol.com


Salynne:
I tried this locally when I 'fixed' it originally and it didn't actually work. Did you test this at all?

steffengy:
yes and it worked. hmm crazy...

Salynne:
Yup confirmed to work. Agreed on the crazy I swear I tried it this way. Must not have clean/packaged, and got the wrong class files.

PR #1755, opened 2012-03-17
Implemented isInUse Method

steffengy:
Implemented the isInUse Method


Afforess:
No, this is wrong. You instead should check chunk.getObservers().size() > 0

Regular entities can observe a chunk too (none do at the moment, only players), but this is so chunks could potentially remain loaded for vehicles, etc.

PR #1760, opened 2012-03-17
Only load regions (and have those start threads) if we are dealing with ...

Darsstar:
...an unloaded world.

Signed-off-by:Darsstar darsstar@gmail.com

### DIFF
--- a/src/main/java/org/spout/server/SpoutServer.java
+++ b/src/main/java/org/spout/server/SpoutServer.java
@@ -264,8 +264,8 @@ public class SpoutServer extends AsyncManager implements Server {
 	}
 
 	public static void main(String[] args) {
-	
-		
+
+
 		SpoutServer server = new SpoutServer(args);
 		server.start();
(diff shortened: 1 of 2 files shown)


src/main/java/org/spout/server/SpoutServer.java
```diff
@@ -650,6 +650,7 @@ public World loadWorld(String name, WorldGenerator generator) {
 			if (!world.getExecutor().startExecutor()) {
 				throw new IllegalStateException("Unable to start executor for new world");
 			}
+			world.start();
```
Raphfrk, line 653: I wonder if that should be before the .startsExecutor

Probably shouldn't matter

Darsstar, line 653: Personally Region threads running without their World's thread running doesn't make much sense to me.
And I can't find any place where SpoutWorld would need a running region thread.

PR #1808, opened 2012-03-24
Multiply the x, y and z values passed into Region's constructor with Region.EDGE

Darsstar:
Signed-off-by:Darsstar darsstar@gmail.com


Raphfrk:
So, I know we discussed this.  The issue is that the cuboid/cube constructors are supposed to take absolute (i.e. block sized) x, y and z coords rather than the number of regions.

Without this change, .getX(), .getY() and .getZ() won't give the right answer?  I think I saw that bug in action.

Darsstar:
That is correct.

Raphfrk:
Merged this using the auto-merge button, I though that included your name in the commit list (rather than just the comment).  Not sure what happened.

PR #1832, opened 2012-03-28
Fixed confusing error message.

credomane:
Fixed confusing error message. (The error was reporting incorrect information.)
Made the invalid names turn into a consistent hex string rather than an always changing one based on time.

Sister pull request https://github.com/SpoutDev/SpoutAPI/pull/147

### DIFF
--- a/src/main/java/org/spout/server/world/SpoutWorld.java
+++ b/src/main/java/org/spout/server/world/SpoutWorld.java
@@ -137,9 +137,11 @@ public class SpoutWorld extends AsyncManager implements World {
 		this.server = server;
 		this.seed = seed;
 		if (!StringSanitizer.isAlphaNumericUnderscore(name)) {
-			Spout.getGame().getLogger().severe("World name " + name + " is not valid, using " + StringSanitizer.asHex(name) + " instead");
+			this.name = StringSanitizer.asHex(name);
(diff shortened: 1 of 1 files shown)


DDoS:
This doesn't set the name to the new hex version, after logging, so the name remains the same after all.

credomane:
Oh darn it. What I get for rushing this at work.

credomane:
Ok apparently that commit got added auto magically? O_o
Anyhow this should be correct, now.

DDoS:
I think you should also squash the commits into one and sign it off.

credomane:
I'd love to but my skills with git aren't good enough squash commits nor do I really understand what this signing-off is. I see it on many of the commits. It doesn't make sense to me because the commit already has my name and email in it. why put it in the commit message too?

Besides I doubt my way is ideal at all. I was just trying to get attention drawn to it and at the very least provide a working solution even if it isn't that great of one.

Raphfrk:
To squash :

```
git rebase -i HEAD~2
<opens a text editor>
on the 2nd line replace "pick" with "s"
save/close file
<opens another text editor>
you can change your comments here if you want
So, add the Signed off by message
save/close file
git push --force origin master
```

When changing the commit message, I think it works if you just add

"Signed-off-by: <your name> <your email>"

Salynne:
Closing this and associated pull request since it's out of date. Thanks for the notification at least, open another pull if you update this.

PR #1844, opened 2012-04-07
Add /reload command. Addresses SPOUT-6.

aPunch:
This PR simply adds a /reload command to Spout. You can pass a plugin to the command if you just wish to reload one plugin.

Is it missing anything? Could the command have more options? Does more need to be reloaded?

### DIFF
--- a/src/main/java/org/spout/engine/command/AdministrationCommands.java
+++ b/src/main/java/org/spout/engine/command/AdministrationCommands.java
@@ -37,6 +37,7 @@ import org.spout.api.command.annotated.Command;
 import org.spout.api.command.annotated.CommandPermissions;
 import org.spout.api.exception.CommandException;
 import org.spout.api.player.Player;
+import org.spout.api.plugin.Plugin;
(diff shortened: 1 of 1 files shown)


Top-Cat:
You've replaced all the tabs with spaces.

aPunch:
Reverted funky formatting issues. It should be good to go formatting-wise, at least.

Top-Cat:
Next up, wouldn't it be better to use an else instead of an if return?
I have a personal hatrid of if returns

Afforess:
> Next up, wouldn't it be better to use an else instead of an if return?

It depends on how it's used. For instance, if return is better here:

```
if (foo) {
     return bar();
} else {
    return null;
}
```

vs

```
if (foo) {
     return bar();
}
return null;
```

But in this case, the return is breaking the standard flow. So yeah, if else would be nice.

aPunch:
Fixed and rebased. What else needs to be changed? :P

Salynne:
Looks good, thanks.

PR #1863, opened 2012-04-25
SpoutEntity refactor and pitch/yaw update

VictorD:
- Swapped pitch and roll around so that pitch is on X axis as Quaternion docs state
- Removed the pitch, yaw and roll instance variables as they can easily be received from the (cached) values in  the quaternion instead. 
- Refactored debug code into one method to reduce code duplication

In hindsight I should probably have split this into multiple PR... oh well let me know if you want me to change anything!

Signed-off-by: Victor Danell victor.danell@gmail.com


Salynne:
Thanks this looks really good :)

PR #1867, opened 2012-05-03
Replaced three-list system in SpoutScheduler

bergerkiller:

Replaced the old/new/current lists with the new concurrent list version.

### DIFF
--- a/src/main/java/org/spout/engine/world/SpoutBlock.java
+++ b/src/main/java/org/spout/engine/world/SpoutBlock.java
@@ -36,7 +36,6 @@ import org.spout.api.geo.discrete.Point;
 import org.spout.api.material.BlockMaterial;
 import org.spout.api.material.block.BlockFace;
 import org.spout.api.material.source.DataSource;
-import org.spout.api.material.source.MaterialData;
 import org.spout.api.material.source.MaterialSource;
 import org.spout.api.math.Vector3;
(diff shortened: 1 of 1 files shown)


Raphfrk:
It would be better to use atomics for the taskId counter, but granted the current implementation doesn't do that.

Why have a specific region task, why not just use the SpoutTask ?

Also, I could complain that you don't have delays/repeating tasks, but we can add that functionality if required.

bergerkiller:
Yeah @Raphfrk This still needs more work. I want to include a SpoutScheduler in the region instead and give regions a 'getScheduler()'. I couldn't use SpoutTask as I had to use my own task ID and I couldn't access the pulse() function. But keep it open, will eventually get something more useful.

bergerkiller:
Changed the PR a bit to make it more simplistic for now. Regions need a full-fledged scheduler which I rather leave up to you to make @Raphfrk

Raphfrk:
One thing with the current scheduler is that it handles tick management and also async tasks.  The regions schedulers would would just be responsible for handling async task.  Maybe that could be split into 2 parts.

bergerkiller:
@Raphfrk Exactly my thought, which was what I tried to do at some point. I guess we need a 'Scheduler' interface for sync stuff, and a 'MainScheduler' to add async scheduling.

PR #1875, opened 2012-05-11
setBlockController now create an entity if needed

bergerkiller:

If a block controller is being set that not yet has an entity attached, it will first create and spawn the entity on the world.


windy1:
@NinjaZidane Can you pull this pretty plox? :)

PR #1876, opened 2012-05-12
SPOUTAPI-27: Add load/generate options (allow load without generate for mapping)

mikeprimm:
See http://issues.spout.org/browse/SPOUTAPI-27 for details.  Depends upon https://github.com/SpoutDev/SpoutAPI/pull/201.

Signed-off-by: mikeprimm mike@primmhome.com

### DIFF
--- a/src/main/java/org/spout/engine/world/RegionSource.java
+++ b/src/main/java/org/spout/engine/world/RegionSource.java
@@ -32,6 +32,7 @@ import java.util.Iterator;
 import org.spout.api.Spout;
 import org.spout.api.event.world.RegionLoadEvent;
 import org.spout.api.event.world.RegionUnloadEvent;
+import org.spout.api.geo.LoadGenerateOption;
 import org.spout.api.geo.cuboid.Chunk;
 import org.spout.api.geo.cuboid.Region;
 import org.spout.api.util.map.concurrent.TSyncInt21TripleObjectHashMap;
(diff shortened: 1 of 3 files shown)


Salynne:
I'm waiting on this because there is still a bunch of stuff messing with saving/loading in a branch.

mikeprimm:
Sounds good - let me know if/when you need these rebased and/or retested

Afforess:
I hate to be a stickler, but you need to squash your commits and need to signoff on your commit. Signing off indicates that you agree to the terms of our license.

mikeprimm:
All good - I'll squash the commits.  What exactly do I need to do to "sign off"?  (I'm fine with doing it - just want to know the process)

Salynne:
http://forums.spout.org/threads/pull-request-guidelines.2922/

mikeprimm:
Cool - thanks for pointer.  Should be all set now, yes?

windy1:
The commit is technically supposed to be signed off :p

mikeprimm:
Was working off what I saw on every other signed pull - didn't see anything on folks commits

Zidane:
Thanks a bunch @mikeprimm

Raphfrk:
Not really a big deal, but it could make things clearer.

Look at how SaveState works in SpoutChunk

https://github.com/SpoutDev/Spout/blob/master/src/main/java/org/spout/engine/world/SpoutChunk.java

This is an enum that has return methods.  Having it as a separate enum is probably better than the way I did it though.

Anyway, you could have 

loadOpt.shouldLoad() - true if the chunk/region should be loaded, if not in memory
loadOpt.shouldGenerate() - true if the chunk/region should be generated, if non-existent

This is better than using ==

mikeprimm:
@Raphfrk - That sounds cool - I'll mod it up tonight, and re-merge the changes

Raphfrk:
@mikeprimm 

Great.  A nice thing about it is that it makes it easier to add extra fields to the enum without breaking code.

PR #1881, opened 2012-05-17
Fixed block updating failures

bergerkiller:

Fixed the failures in SpoutChunk and Column which caused some blocks to never update.
See also:
https://github.com/SpoutDev/SpoutAPI/pull/212

### DIFF
--- a/src/main/java/org/spout/engine/world/SpoutChunk.java
+++ b/src/main/java/org/spout/engine/world/SpoutChunk.java
@@ -600,8 +600,8 @@ public class SpoutChunk extends Chunk {
 		return blockStore.isDirtyOverflow();
 	}
 
-	public Block getDirtyBlock(int i) {
-		return blockStore.getDirtyBlock(i, this.getWorld());
+	protected Vector3 getDirtyBlock(int i) {
+		return this.blockStore.getDirtyBlock(i);
 	}
 
 	public void resetDirtyArrays() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/spout/engine/world/SpoutChunk.java
```diff
@@ -946,6 +941,14 @@ public Block getBlock(Vector3 position, Source source) {
 	}
 
 	@Override
+	public Block getBlock(int x, int y, int z, Source source) {
+		x = (x & CHUNK_SIZE_BITS) + this.getX() * CHUNK_SIZE;
```
Raphfrk: I think this should be (and the others need to be changed too):

```
x = (x & (~BASE_MASK)) + (getX() << CHUNK_SIZE_BITS);
```

CHUNK_SIZE_BITS means the number of bits in the chunk size and is equal to 4.

bergerkiller: @Raphfrk You are right about that, fixed it.

Raphfrk:
How does this interact with players.  Player observers are handled by network synchronizer.

Presumably, observers cause chunks to be loaded?  So care really needs to be taken with these.

LunNova:
> Player observers are handled by network synchronizer.

Obviously not correctly - it didn't work. :(

bergerkiller:
@Raphfrk That only handled the sending of chunks to the players. The observe bit (which changes in what chunk are visible to what players, whether entity is visible, block changes and others) were never set. And I see no effective way to do that from a synchronizer.

Afforess:
> How does this interact with players. Player observers are handled by network synchronizer.

Yes, I wanted to discuss this with you, I don't think they are handled correctly. I've confirmed that any chunks an observer is not directly inside of is unloaded after the timeout. 

However, I already had a fix similar to this pull request to this in my local repo (http://spout.in/280c7e6), so I am closing, but the topic needs further discussion.

Raphfrk:
There are observers and players.  When a player is viewing a chunk, it adds the player as an observer of the chunk.

However, setting an entity as an observer, just adds the entity to the current chunk.  It doesn't even support changing chunk properly.

Unless it has changed, the network synchronizer handled adding the player to chunks the player was observing.  In fact, that was what observers was meant to mean, it was the list of players who were viewing the chunk.

PR #1884, opened 2012-05-22
Add chunk load/unload/populate/update events 

mikeprimm:
Add events for various chunk lifecycle points:
1) Chunk loaded, including when chunk is generated
2) Chunk populated
3) Chunk unloaded
4) Chunk updated (based on reporting dirty block coordinates or whole chunk update, as used in network synchronizer).

All are non-cancellable, monitor style events, useful for monitoring world data changes without interfering with them (whether driven by player action or by plugins).

Codependent with SpoutAPI pull https://github.com/SpoutDev/SpoutAPI/pull/218.

Signed-off-by: Mike Primm mike@primmhome.com

### DIFF
--- a/src/main/java/org/spout/engine/world/SpoutRegion.java
+++ b/src/main/java/org/spout/engine/world/SpoutRegion.java
@@ -28,6 +28,7 @@ package org.spout.engine.world;
 import java.io.File;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -49,6 +50,10 @@ import org.spout.api.entity.BlockController;
 import org.spout.api.entity.Controller;
 import org.spout.api.entity.Entity;
 import org.spout.api.entity.PlayerController;
+import org.spout.api.event.chunk.ChunkLoadEvent;
+import org.spout.api.event.chunk.ChunkPopulateEvent;
+import org.spout.api.event.chunk.ChunkUnloadEvent;
+import org.spout.api.event.chunk.ChunkUpdatedEvent;
 import org.spout.api.generator.WorldGenerator;
 import org.spout.api.generator.biome.Biome;
 import org.spout.api.generator.biome.BiomeManager;
@@ -219,13 +224,14 @@ public class SpoutRegion extends Region {
 			AtomicReference<SpoutChunk> ref = chunks[x][y][z];
 
 			boolean success = false;
-
+			boolean generated = false;
 			SpoutChunk newChunk = WorldFiles.loadChunk(this, x, y, z, this.getChunkInputStream(x, y, z));
 			if (newChunk == null) {
 				if (!loadopt.generateIfNeeded()) {
 					return null;
 				}
 				newChunk = generateChunk(x, y, z);
+				generated = true;
 			}
 
 			while (!success) {
@@ -238,6 +244,9 @@ public class SpoutRegion extends Region {
 						nonPopulatedChunks.add(newChunk);
 					}
 					runningChunksQueue.add(newChunk);
+
+					Spout.getEventManager().callDelayedEvent(new ChunkLoadEvent(newChunk, generated));
+
 					return newChunk;
 				} else {
 					newChunk.deregisterFromColumn(false);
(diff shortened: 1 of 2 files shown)


src/main/java/org/spout/engine/world/SpoutRegion.java
```diff
@@ -624,6 +646,35 @@ private void syncChunkToPlayers(SpoutChunk chunk, Entity entity) {
...
+		}
+		if (chunk.isNewlyLoaded()) {	// If newly loaded, don't report initial touches
+			return;
+		}
+		ChunkUpdatedEvent evt;
+		if (chunk.isDirtyOverflow()) {	/* If overflow, notify for whole chunk */
+			evt = new ChunkUpdatedEvent(chunk, null);
+		}
+		else {
+			ArrayList<Vector3> lst = new ArrayList<Vector3>();
```
Raphfrk, line 680: Do we want to make this an unmodifyable list, to prevent plugins altering it.

Making a copy is good, since at least plugins can't break the core server, but maybe they could be protected against each other

src/main/java/org/spout/engine/world/SpoutRegion.java
```diff
@@ -624,6 +646,35 @@ private void syncChunkToPlayers(SpoutChunk chunk, Entity entity) {
 			}
 		}
 	}
+	
+	private void processChunkUpdatedEvent(SpoutChunk chunk) {
+		/* If no listeners, quit */
+		if (ChunkUpdatedEvent.getHandlerList().getRegisteredListeners().length == 0) {
+			return;
+		}
+		if (chunk.isNewlyLoaded()) {	// If newly loaded, don't report initial touches
```
Raphfrk: Initial touches means initialization of the chunk ?

src/main/java/org/spout/engine/world/SpoutRegion.java
mikeprimm, line 680: Can't hurt the server - the Vector3 values generated when they are fetched (they aren't the internal values in the map), so the array and its contents are only for the event (output only)

src/main/java/org/spout/engine/world/SpoutRegion.java
mikeprimm: Just loading the chunk winds up resulting in it being marked as dirty (artifact of the way the data map works - loading in the data array trips off all the dirty block logic), which would result in incorrect reporting of updates that either didn't happen, or are implicit in the fact that the chunk is newly generated (and which is implicit in the isNewBlock=true in the ChunkLoadEvent).

src/main/java/org/spout/engine/world/SpoutRegion.java
Raphfrk, line 680: Right, as I said, it would only be to protected plugins from each other.

src/main/java/org/spout/engine/world/SpoutRegion.java
Raphfrk: However, updates that happen in the first tick would be missed.  Not sure of a clean way to handle it.

mikeprimm: Dunno - I'd considered whether the dirty flags being set during the constructor should just be cleared (on the assumption that any "dirty-ness" at that point is bogus).  But, since it winds up causing any observers that jump right in to get a full chunk update, I was concerned that it might be needed-if-not-a-bit-questionable.

src/main/java/org/spout/engine/world/SpoutRegion.java
mikeprimm, line 680: OK - given how unprotected things are in general, I hadn't considered defending the events against hostile or ill-behaved mods as a priority....

Raphfrk, line 680: Right.

src/main/java/org/spout/engine/world/SpoutRegion.java
Raphfrk: We could clear the dirty flag at the end of the constructor, assuming the updates are happening there.

mikeprimm: I'll give that a go, as well as looking at the possibility of a cancellable unload event.  The main question on the cancellable unload is whether or not to fire it when an unconditional unload is occurring (that is, if the unload() call is being made directly, or in conjunction with another un-cancellable situation - region.unload, world.unload, server shutdown).  Alternately, the event could be 'cancellable' conditionally - it could include a flag indicating that the specific event is actually NOT cancellable, in the context that it is being emitted.

src/main/java/org/spout/engine/world/SpoutRegion.java
mikeprimm, line 680: I'll probably switch it to something contained (with a count method and a get-by-index method).

mikeprimm:
Anything more you guys think I should do here?  Dropped the ChunkReap event, and made ChunkUnload cancellable, as requested.

Raphfrk:
I think it's good, needs one final pass (and a manual merge)

PR #1894, opened 2012-05-27
Ran FindBugs on Spout

TomyLobo:
The changes from this last commit should not be pulled, but inspected by the respective developers.

I didn't sign it off to hint at this.

I also made one for SpoutAPI: https://github.com/SpoutDev/SpoutAPI/pull/230
and one for Vanilla: https://github.com/VanillaDev/Vanilla/pull/213

### DIFF
--- a/src/main/java/org/spout/engine/SpoutEngine.java
+++ b/src/main/java/org/spout/engine/SpoutEngine.java
@@ -300,6 +300,7 @@ public class SpoutEngine extends AsyncManager implements Engine {
 
 	@Override
 	public String[] getAllAddresses() {
+		// FIXME: This is equivalent to "return null;" right now, since the field is never ever written.
 		return allAddresses;
 	}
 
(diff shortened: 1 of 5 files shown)


windy1:
These should be squashed into one commit.

TomyLobo:
no, I want the fixes to be looked at by the developers
Commits 2 and 3 are things that really shouldn't happen to anyone trusted with such a project.
Commit 2 is the result of someone being unaware of object array cast rules
Commit 3 is apparently the result of someone uncritically applying quickfixes
Commit 4 is just an oversight
Commit 5 needs to be looked at by the developers and then discarded (it only adds comments)

TomyLobo:
ok, that leaves only commit #5, to be inspected by some people :)

Salynne:
Closing these, and making some probably permanent issues for the repos as a reminder to keep up to date on this.

http://issues.spout.org/browse/SPOUT-66

PR #1897, opened 2012-05-28
Implemented /Help Command

fallensk8ur:
Lists all commands registered with the server.
Signed-off-by: Fallensk8ur fallensk8ur1234@aol.com


Salynne:
Not really what we're looking for. If we add something like this it will do it dynamically.

PR #1898, opened 2012-05-28
Added populated check

bergerkiller:
- To prevent a chunk being sent before it is populated.


Raphfrk:
I don't think this is correct.  We still want to send updates if they happen to chunks pre-population?

Zidane:
I think lighting needs to be calculated after all population that occurs in a chunk to prevent "double calculation" @Raphfrk

Raphfrk:
Ok, but this blocks sending updates to the client completely.

Zidane:
@Raphfrk hmmm

What do we do then? :p I noticed stuff like redstone is still working fine :S.

Raphfrk:
I mean, it blocks sending any updates to non-populated chunks.

It isn't really critical, since presumably, populating the chunk will cause it to update anyway.

PR #1900, opened 2012-05-30
Cleanup of SpoutEngine version 2 - Zidane's formatter broke.

LunNova:
Since Zidane appears to be unable to reformat, done this instead.

Signed-off-by: Ross Allan rallanpcl@gmail.com


Salynne:
I told him not to anyway, and I also don't accept formatting pull requests unless there is some really good reason. @RoyAwesome can cleanup his own code.

PR #1901, opened 2012-05-31
Added check when sending chunk for pending light

bergerkiller:

It was sending chunks which were still undergoing lighting calculations, which caused the client to lag like crazy when entering newly populated chunks. This should deal with this problem.


src/main/java/org/spout/engine/world/SpoutChunk.java
```diff
@@ -139,7 +139,7 @@
 	 * Every time a change is committed the value is set to 0. The region will increment it as well.
 	 */
 	protected final AtomicInteger lightingCounter = new AtomicInteger(-1);
-
+
```
md-5: git-fail

md-5:
look ma, I helped! :P

PR #1909, opened 2012-06-07
Prevent cascading populators...

DDoS:
...which would cause 100% RAM and CPU usage (and often server crashes and "out of heap space exceptions").

This would happen when a populator/decorator would overflow on another chunk (think of a tree, on the edge of a chunk, some leaves could end up on the neighboring chunk), and cause this chunk to be loaded and populated, and possibly overflowing too, causing an infinite chain.

This PR adds an observer check before unforced population, making sure the chunk is being observed, so that cascading will stop when players can't see the to-be-populated chunks anymore.

Also, some random formatting, which I have no idea where it came from... The actual change is at line 823.

Signed-off-by: Aleksi Sapon QcTechs@gmail.com


Raphfrk:
95% of the pull doesn't change anything.

I think the only functional change is:

```
if (!force && observers.getLive().size() <= 0 && observers.get().size() <= 0) {
    return;
}
```

Can you checkout the old file and just add your few lines.

As for the pull itself, I thought we did that already.  The change is good.

DDoS:
Yes, I know most of it doesn't change anything, I think it's line ending formatting git added when I commited the changes... since I don't remember ever using the formatting command in my IDE. Since this is automatic stuff, I don't think I can remove those changes without changing my line endings setting...

DDoS:
Closing the PR since the fix apparently doesn't work well with more than one person online.

DDoS:
I was not able to reproduce the reported problem, and I think it might just have been mistake; the smooth populator occasionally fails at smoothing which leaves the terrain looking un-smoothed (problem which I fixed in my last commit).

And so I'm reopening this PR.

Raphfrk:
I added this as part of another commit.

https://github.com/SpoutDev/Spout/commit/9aedcb53d44f67722a4815f71755da76d8341101

DDoS:
Thanks!

PR #1913, opened 2012-06-12
User friendly error on port bind error.

Wouterdek:
This gives a more user-friendly error when a user tries to run multiple individual servers on the same port.


windy1:
Formatting is off, you need to use tabs not spaces. :)

lukespragg:
It also needs to be about the port, not address really. Multiple servers can run on the same address, but not the same port. There are quite a few issues with formatting as Windwaker said as well. Should be spacing around the + signs and inside each { as well as tabs, not spaces for indentation. Commits also need to be signed off. See http://spout.in/prguide

Afforess:
Fixed and pulled here: https://github.com/SpoutDev/Spout/pull/1925

PR #1927, opened 2012-06-29
Cleanup warnings

tips48:
I cleaned up most of the warnings in all the Spout projects. It was mainly imports, but there were some issues with not closing resources and type casting.
Enjoy!


src/main/java/org/spout/engine/listener/SpoutProxyConnectListener.java
```diff
@@ -41,7 +40,8 @@
 
 public class SpoutProxyConnectListener implements ChannelFutureListener {
 	
-	private final Engine engine;
+	@SuppressWarnings("unused")
+        private final Engine engine;
```
Raphfrk, line 44: Is that correct formatting?

Afforess, line 44: Nope. engine should not be indented.

tips48, line 44: My client derped whenever I added a @SurpressWarnings and indented the next line..I'll fix next time I'm free

src/main/java/org/spout/engine/batcher/GuiRenderer.java
```diff
@@ -50,12 +47,14 @@ public static void init(){
 		
 	}
 	
-	private static void TexturedQuad(float x, float y, float width, float height){
+	@SuppressWarnings("unused")
+        private static void TexturedQuad(float x, float y, float width, float height){
```
olliestanley, line 51: Just so you don't miss it, formatting here as well

PR #1929, opened 2012-07-02
Fixed infinite loop issue

bergerkiller:

Triggering the same block twice would cause 2 dynamic update entries per block. A while loop was not properly going to the next element.


Raphfrk:
Good catch

PR #1940, opened 2012-07-16
Disabled physics when populating

bergerkiller:

Executing physics for every single block is too slow for the world physics to handle. To boost performance, we should rely on the populator doing it's work well without physics being needed at all.

Dynamic material resets are still called, as blocks like cactuses and fire rely on these updates to function. Else they wouldn't grow.

### DIFF
--- a/src/main/java/org/spout/engine/world/SpoutChunk.java
+++ b/src/main/java/org/spout/engine/world/SpoutChunk.java
@@ -1607,6 +1607,7 @@ public class SpoutChunk extends Chunk {
 	}
 
 	private void blockChanged(int x, int y, int z, BlockMaterial newMaterial, short newData, BlockMaterial oldMaterial, short oldData, Source source) {
+		// Handle onPlacement for dynamic materials
 		if (newMaterial instanceof DynamicMaterial) {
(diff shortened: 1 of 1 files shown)


Spouty:
**[AUTO-GENERATED COMMENT]**

Please make sure this pull request follows the [pull request guidelines](http://spout.in/prguide) in order to ensure your pull request is accepted.

src/main/java/org/spout/engine/world/SpoutChunk.java
```diff
@@ -1617,14 +1618,18 @@ private void blockChanged(int x, int y, int z, BlockMaterial newMaterial, short
...
+		if (this.isPopulated()) {
+			EffectRange physicsRange = newMaterial.getPhysicsRange(newData);
+			queueBlockPhysics(x, y, z, physicsRange, oldMaterial, source);
+			if (newMaterial != oldMaterial) {
+				EffectRange destroyRange = oldMaterial.getDestroyRange(oldData);
+				if (destroyRange != physicsRange) {
+					queueBlockPhysics(x, y, z, destroyRange, oldMaterial, source);
+				}
 			}
+			return;
```
Afforess: Why the return?

PR #1943, opened 2012-07-22
Removed Block getSubMaterial

bergerkiller:

Removed getSubMaterial (reason: API change)

https://github.com/SpoutDev/SpoutAPI/pull/300


src/main/java/org/spout/engine/world/SpoutBlock.java
```diff
@@ -253,11 +253,6 @@ public Source getSource() {
 	}
 
 	@Override
-	public BlockMaterial getSubMaterial() {
-		return this.getMaterial().getSubMaterial(this.getData());
-	}
-
-	@Override
```
Raphfrk, line 260: why ?

PR #1950, opened 2012-07-27
Performance improvements for the Block constructor

bergerkiller:

More than often translating the Block results in a Block in the same chunk as the previous. This PR checks if the new block still resides in the chunk given (also a sort of safety feature) and also passes the current chunk into the block constructor when translating.

I changed the 'get chunk' call in the chunk-free constructor to null to make sure that, when you don't need a chunk, no chunk is obtained. For example, you translate a block two times and discard the previous block.

### DIFF
--- a/src/main/java/org/spout/engine/world/SpoutBlock.java
+++ b/src/main/java/org/spout/engine/world/SpoutBlock.java
@@ -57,9 +57,9 @@ public class SpoutBlock implements Block {
 	private final WeakReference<World> world;
 	private final Source source;
 	private final AtomicReference<WeakReference<Chunk>> chunk;
-	
+
 	public SpoutBlock(World world, int x, int y, int z, Source source) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/engine/world/SpoutBlock.java
```diff
@@ -71,9 +71,12 @@ protected SpoutBlock(World world, int x, int y, int z, Chunk chunk, Source sourc
 		this.z = z;
 		this.world = new WeakReference<World>(world);
 		this.source = source;
+		if (chunk != null && !chunk.containsBlock(this.x, this.y, this.z)) {
+			chunk = null; // chunk does not contain this Block, invalidate
```
Afforess, line 75: Shouldn't you look up the chunk here?

bergerkiller, line 75: No, the Chunk weak reference is allowed to contain null. It would then get the chunk from the world when getChunk is called.

PR #1953, opened 2012-07-29
Fixed block controllers getting mapped to points instead of vector3

bergerkiller:

Fixes all the controller related problems.


macalinao:
Holy sh*t good job

PR #1955, opened 2012-08-01
Moved ChunkMesh loading logic to an update() method and improved memory ...

macalinao:
...usage.

Signed-off-by: Ian Macalinao ianmacalinao@gmail.com


macalinao:
Didn't know about the feature lock, I'll work on this on my own for the next 3 days.

PR #1966, opened 2012-08-13
Moved Spout built-in protocol stuff to Spout.

macalinao:
Signed-off-by: Ian Macalinao ianmacalinao@gmail.com

https://github.com/SpoutDev/SpoutAPI/pull/330


Raphfrk:
Why

Afforess:
I echo @Raphfrk's question.

Zidane:
@Raphfrk @Afforess 

Simple, for security purposes we shouldn't expose the built-in protocol to the API. If we design the system well, then messages for core functions such as AddEntity or WorldChange would never need to be sent manually or constructed differently.

If a plugin needs to send their own messages, do the same as Vanilla. Construct their own protocol with their own NetworkSynchronizer. My only personal thing is that a plugin should be able to add their own possible messages into Spout Protocol  so users don't have to connect to another port like Vanilla (unless that is already possible).

PR #1970, opened 2012-08-18
Made lighting a tick stage task

bergerkiller:

I simplistic attempt to turn the lighting thread into a world stage. I was unable to add a new world stage for it, because there was no setting for it in the world (the constructor one had a TODO and didn't work, tested)

The operations succeed, but chunks are sent too fast, resulting in incomplete chunks being received. This may need a separate fix. (was already happening prior to this change)

### DIFF
--- a/src/main/java/org/spout/engine/world/SpoutWorld.java
+++ b/src/main/java/org/spout/engine/world/SpoutWorld.java
@@ -218,7 +218,6 @@ public class SpoutWorld extends AsyncManager implements World {
 		this.hashcode = new HashCodeBuilder(27, 971).append(uid).toHashCode();
 
 		this.lightingManager = new SpoutWorldLighting(this);
-		this.lightingManager.start();
 
 		parallelTaskManager = new SpoutParallelTaskManager(engine.getScheduler(), this);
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/spout/engine/world/SpoutWorld.java
```diff
@@ -606,13 +605,14 @@ public void copySnapshotRun() throws InterruptedException {
 
 	@Override
 	public void startTickRun(int stage, long delta) throws InterruptedException {
-		if (stage == 0) {
-			age.set(age.get() + delta);
-		}
 		switch (stage) {
 			case 0: {
+				age.set(age.get() + delta);
```
Raphfrk, line 610: This would mean that it executes at the same time as anything else during the tick stage.

Also, it still makes it single threaded.  Ideally, it would run sometime near when physics is run.

Maybe, it would be run after physics/dynamic updates and before finalize.

It would be best to have its own tick stage.

What would be involved in making it work on a region basis?

It could operate like physics, where mostly it is region based, and then does global updates separately.

Salynne:
Closing for the moment, this can be updated into the tick stage for this that @Raphfrk added

PR #1973, opened 2012-08-20
Added SpoutPlayer positional saving.

R-T-B:
Uses saveEntity() like all good code should.  Saves in data/players/player.dat
Signed-off-by: Raymond Beehler 21GunSoftware@comcast.net

### DIFF
--- a/src/main/java/org/spout/engine/SpoutEngine.java
+++ b/src/main/java/org/spout/engine/SpoutEngine.java
@@ -77,6 +77,7 @@ import org.spout.api.generator.biome.BiomeRegistry;
 import org.spout.api.geo.World;
 import org.spout.api.geo.cuboid.Region;
 import org.spout.api.geo.discrete.Point;
+import org.spout.api.geo.discrete.Transform;
 import org.spout.api.inventory.CommonRecipeManager;
 import org.spout.api.inventory.RecipeManager;
 import org.spout.api.io.store.simple.BinaryFileStore;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spout/engine/SpoutEngine.java
```diff
@@ -380,6 +381,13 @@ public File getConfigFolder() {
 
 	@Override
 	public File getDataFolder() {
+		if (!dataDirectory.exists()) {
+			dataDirectory.mkdirs();			
+		}
+		File playerDirectory = new File(dataDirectory.toString() + File.separator + "players");
+		if (!playerDirectory.exists()) {
+		playerDirectory.mkdirs();
```
Darsstar: not indented

src/main/java/org/spout/engine/SpoutEngine.java
```diff
@@ -763,28 +771,22 @@ public Entity getEntity(UUID uid) {
...
-
-		while (true) {
-			player = onlinePlayers.getLive().get(playerName);
-
-			if (player != null) {
-				if (!player.connect(session, getDefaultWorld().getSpawnPoint())) {
-					return null;
-				}
-				break;
+			try {
```
Darsstar: try and catch blocks are indented one level to much.

the rest of the method except for the return satement is indented even more.

src/main/java/org/spout/engine/filesystem/WorldFiles.java
```diff
@@ -101,7 +105,56 @@
 	private static final byte ENTITY_VERSION = 1;
 	private static final byte CHUNK_VERSION = 1;
 	private static final int COLUMN_VERSION = 4;
-
+	public static void savePlayerData(List<SpoutPlayer> Players) {
```
Darsstar: add the newline back, add newlines between methods and replace the spaces (savePlayerData(SpoutPlayer player) only) by tabs.

PR #1974, opened 2012-08-26
Altered region entity manager to use message buffer

bergerkiller:

Also prevents updating the same entity twice, and generates one list of messages to fill and re-use as buffer.

All pulls:
https://github.com/VanillaDev/Vanilla/pull/299
https://github.com/SpoutDev/SpoutAPI/pull/340
https://github.com/SpoutDev/Spout/pull/1974

### DIFF
--- a/src/main/java/org/spout/engine/entity/RegionEntityManager.java
+++ b/src/main/java/org/spout/engine/entity/RegionEntityManager.java
@@ -27,15 +27,10 @@
 package org.spout.engine.entity;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import com.google.common.collect.Lists;
-
-import org.spout.api.entity.Entity;
 import org.spout.api.entity.Player;
 import org.spout.api.math.MathHelper;
 import org.spout.api.protocol.NetworkSynchronizer;
@@ -93,7 +88,7 @@ public final class RegionEntityManager extends EntityManager {
 	}
 
 	public List<Player> getPlayers() {
-		Map map = players.get();
+		Map<?, ?> map = players.get();
 		if (map == null) {
 			return Collections.emptyList();
 		}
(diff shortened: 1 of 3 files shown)


src/main/java/org/spout/engine/entity/RegionEntityManager.java
```diff
@@ -118,32 +115,28 @@ public void syncEntities() {
 			if (entitiesPerPlayer == null) {
 				entitiesPerPlayer = new ArrayList<SpoutEntity>();
 			}
+			boolean spawn, destroy, update;
 			for (SpoutEntity entity : getAll()) {
-				if (entity.equals(player)) {
+				if (entity == null || entity.equals(player)) {
```
Afforess: entity should never be null here. If it is, that points to other problems.

src/main/java/org/spout/engine/protocol/builtin/SpoutEntityProtocol.java
```diff
@@ -46,26 +45,25 @@ protected SpoutEntityProtocol() {
 		super();
 	}
 
-	public Message[] getSpawnMessage(Entity entity) {
-		return new Message[] {new AddEntityMessage(entity.getId(), entity.getController().getType(), entity.getTransform())};
+	@Override
+	public void getSpawnMessages(Entity entity, List<Message> messages) {
```
Afforess: A couple of comments here:
- the method is misnamed. It shouldn't be "get" because it doesn't return anything. AddSpawnMessages maybe?
- It's scary that the list is exposed. I understand it's for performance reasons, but theoretically couldn't someone accidently do a .clear() and break the server? Do we need a special subclass of list that hides those methods? Something else?

bergerkiller:
The List<Message>.clear/remove problem still exists. What to do?:
- Make a separate collection that only adds? (it still has to clear though...)
- Let the protocol send messages directly and don't use a List

Spouty:
**[AUTO-GENERATED COMMENT]**

PR #1976, opened 2012-08-27
Update for ban changes and added ban commands

windy1:
SpoutAPI PR: https://github.com/SpoutDev/SpoutAPI/pull/342
- Added 'ban', 'unban', 'banip', and 'unbanip' commands

Signed-off-by: Walker Crouse walkercrouse@hotmail.com


dredhorse:
as far as I can see no single Event was implemented, I guess I will have to redo my pull request and implement that again.

windy1:
@dredhorse Sorry, when was it ever stated that this PR had anything to do with events?

dredhorse:
nowhere, but I added a pull request which implements them for those commands 21 days ago.

also I would assume that if you write new code that you than try to implement also the needed functions into the new code. I wonder why we have a kickbanevent if it is not used.

sorry but I was pissed off yesterday after being awake for 20 hours that I wait for feedback why my pr isn't pulled and 2 PR's are just pulled without any discussion.

Also afaik you didn't really clean up the code in the area that there is also a ban list in the spout config, which is never used anywhere.

windy1:
@dredhorse While I can understand why you are upset, this wasn't about cleanup, it was about functionality change.

dredhorse:
well my PR also had the additional commands, you could have pulled it and than added the UN**\* commands..

whatever, just redoing my PR.. hope I don't make big fubars while merging

PR #1983, opened 2012-09-03
Pull Request so this can be looked at first - Reusing Player objects during addPlayer

Zidane:


src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -70,7 +70,7 @@
 	private final AtomicReference<Transform> transformLive;
 	private final AtomicBoolean deadLive = new AtomicBoolean(false);
 	private final AtomicBoolean observerLive = new AtomicBoolean(false);
-	private final AtomicInteger id = new AtomicInteger();
+	protected final AtomicInteger id = new AtomicInteger();
```
Afforess: I'd prefer this be kept private and you add a resetId() to SpoutEntity instead. ID is a component of entities, not players.

src/main/java/org/spout/engine/SpoutEngine.java
```diff
@@ -708,17 +708,28 @@ public Entity getEntity(UUID uid) {
...
 		if (player == null) {
-			//generate a player at default spawn, since we don't have a player .dat file
-			player = new SpoutPlayer(playerName, getDefaultWorld().getSpawnPoint(), session, this, viewDistance);
+			//Attempt to locate and load player data from file
+			player = WorldFiles.loadPlayerData(playerName, session);
+			//Player has no saved data, create a new object for them
+			if (player == null) {
+				player = new SpoutPlayer(playerName, getDefaultWorld().getSpawnPoint(), session, this, viewDistance);
+			}
+			onlinePlayers.put(playerName, player);
```
Raphfrk, line 720: This could potentially fail if a player connects twice from 2 different clients at the same instant.

It is low probability though.

Anyway, the code should be something like

```
SpoutPlayer player = onlinePlayers.getLive().get(playerName);
//Player is joining the server for the first time
if (player == null) {
    //Attempt to locate and load player data from file
    player = WorldFiles.loadPlayerData(playerName, session);
    //Player has no saved data, create a new object for them
    if (player == null) {
        player = new SpoutPlayer(playerName, getDefaultWorld().getSpawnPoint(), session, this, viewDistance);
    }
    Player oldPlayer = onlinePlayers.putIfAbsent(playerName, player);
    if (oldPlayer != null) {
        // Player object was created by another thread between the .get() call and the .putIfAbsent() call 
        player = oldPlayer;
        // Does calling new SpoutPlayer() have any side effects or can the object just be discarded and GC-ed?
    }
}
// Can call this section always, it does no harm for new Player objects since they have null sessions anyway
player.resetId();
if (player.getSession() != null) {
    player.kick("Logged in from another client");
}
```

The kick part isn't technically threadsafe either.  You kind of want getAndSet() on the session, so that only 1 thread will send the kick message.  Maybe .connect() could return the old session or something.

Zidane, line 720: I reverted this as it was causing weird session connection issues @Raphfrk so perhaps we need to take another look at things. I half expect clients not receiving the session during connect is resulting from a threading issue.

Raphfrk, line 720: hmm, it should only really have caused an issue if someone was connecting over and over and even then, it would probably be ok.

Were you just connecting with one client?

Zidane, line 720: Yeah, it was random but apparently "reproduceable" if you teleport to a new world then log out you couldn't log back in.

PR #1992, opened 2012-10-12
 World renderer in gl11 and gl20

Karang:
fix the world selection, fix gl11 renderer.

### DIFF
--- a/src/main/java/org/spout/engine/SpoutClient.java
+++ b/src/main/java/org/spout/engine/SpoutClient.java
@@ -291,10 +291,10 @@ public class SpoutClient extends SpoutEngine implements Client {
 		if (!Keyboard.isCreated())
 			return;
 
-		boolean keyUp = Keyboard.isKeyDown(Keyboard.KEY_UP) || Keyboard.isKeyDown(Keyboard.getKeyIndex(SpoutInputConfiguration.FORWARD.getString()));
-        boolean keyDown = Keyboard.isKeyDown(Keyboard.KEY_DOWN) || Keyboard.isKeyDown(Keyboard.getKeyIndex(SpoutInputConfiguration.BACKWARD.getString()));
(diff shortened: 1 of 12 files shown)


src/main/java/org/spout/engine/SpoutClient.java
```diff
@@ -390,7 +390,7 @@ public void stopEngine() {
 		super.stop(stopMessage);
 	}
 
-	@Override
+	/*@Override //Because there is a conflict when the spout engine tries to load the world
```
zml2008, line 393: Then fix it in the spout engine, not here.

src/main/java/org/spout/engine/SpoutEngine.java
```diff
@@ -656,7 +656,7 @@ public World getDefaultWorld() {
 			return null;
 		}
 
-		final World first = loadedWorlds.values().iterator().next();
```
zml2008, line 659: I don't  see the point of this.

src/main/java/org/spout/engine/SpoutClient.java
Karang, line 393: the issue is that the spout engine called the getWorld() function but because the instance of the engine used by the client is a spoutClient wich is a subclass of spoutEngine, so its this function that is called, but the function that call it precisly the function that set the activeworld so it'll always be null before that. And it doesn't make any sens. Also this function have no other interrest for now so it can be just commented.

Spouty:
**[AUTO-GENERATED COMMENT]**

src/main/java/org/spout/engine/SpoutEngine.java
Karang, line 659: yes that was a mistake because we are 2 working on the same project via saros and i tried to clean some code but didn't write the same line of code, but you have to admit it's pretty close :D

PR #1996, opened 2012-10-16
World textures + GL11 input and textures fix

Karang:


RoyAwesome:
FYI, not going to pull this just because it was made literally 10 minutes before I exposed texturing to the API

Materials now have a String texturePath and a Rectangle that is their texcoords.  Rectangle now has a helper method to get a set of texcoords from a given texture atlas.  

Finally, not pulling it because terrain.png is in vanilla ;)

PR #1999, opened 2012-10-17
Replace hack input by use of InputCommand & PlayerInputState

ghost:


narrowtux:
Why did you remove the usage of our Keyboard class?
It solves the sticky-key problem and as I can see, SpoutInput doesn't keep track of all keys so it wouldn't be suitable for working around the bug.

Also, please do more descriptive commit messages.

narrowtux:
As in, What does this pull request do?

ghost:
As i say on forum, sticky key can be clear in PlayerInputState, just we need a func to clear flag.

This pull request replaces the hack movements client by the use of InputCommand and PlayerInputState.

SpoutInput is poll before each render frame, each command bind with a key press or release is send with player as source. The inputcommands modify the PlayerInputState, and finaly the playerinputstate is used to actualize player position before rendering.

PR #2005, opened 2012-10-19
Binding gui with the api. 

Karang:


src/main/java/org/spout/engine/SpoutClient.java
```diff
@@ -217,6 +221,16 @@ public void start(boolean checkWorlds) {
 		activePlayer = new SpoutClientPlayer("Spouty", loc, SpoutConfiguration.VIEW_DISTANCE.getInt() * Chunk.BLOCKS.SIZE);
 		activeCamera = activePlayer.add(CameraComponent.class);
 		mainScreen = new Screen();
+		
+		/*Widget txtWidget = new Widget(); // Waiting for the NPE in Widgets to be corrected
+		txtWidget.add(SpoutLabelComponent.class);
```
Zidane, line 226: add returns either the component if already existant or the component created. No sense in doing a get on further lines :p.

PR #2006, opened 2012-10-20
Use scheduler to generate mesh with chunk, use renderTaskQueue to generate display list.

ghost:


src/main/java/org/spout/engine/mesh/ChunkMesh.java
```diff
@@ -53,11 +54,12 @@
 	private final Chunk chunk;
 	private ChunkSnapshotModel chunkModel;
 	private ChunkSnapshot center;
+	public Semaphore lock = new Semaphore(1);
 
 	/**
 	 * Private constructor.
```
olliestanley: JavaDoc should probably be changed :P

src/main/java/org/spout/engine/batcher/ChunkMeshBatch.java
```diff
@@ -48,29 +51,39 @@
...
 		int id = 0;
 		for (int i = baseX; i < baseX + SIZE_X; i++) {
 			for (int j = baseY; j < baseY + SIZE_Y; j++) {
 				for (int k = baseZ; k < baseZ + SIZE_Z; k++) {
 					SpoutSnapshotLock lock = (SpoutSnapshotLock) Spout.getEngine().getScheduler().getSnapshotLock();
+
+					lock.coreReadLock("Generate mesh");
+					Chunk chunk = world.getChunk(i, j, k);
+					meshes[id] = new ChunkMesh(chunk);
+					lock.coreReadUnlock("Generate mesh");
```
Raphfrk: This should use the try ... finally structure, as with the commented out portion.

lock.coreReadLock(...);
try {
.....
} finally {
    lock.coreReadUnlock(....);
}

This guarantees that the lock is released, even if an exception is thrown.  With locks, it is much better to be safe than sorry.

Raphfrk: Also, there is a snapshot system already built into SpoutChunk for the renderer.

Raphfrk: https://github.com/SpoutDev/Spout/blob/master/src/main/java/org/spout/engine/world/SpoutChunk.java#L874

So, something like

((SpoutChunk) chunk).getFutureSnapshot(..., ..., ..., true);

You can only have one future pending per chunk.  So, make sure you wait until all current futures are free before calling any more.

The method returns null if there has been no changes since the last call.

The futures are processed at the end of the current tick.  This eliminates the need to use snapshot locking at all.

However, you do need to be able to handle the future queue.

ghost: I will try to use it, but the actual code is complicated.
i have discover that this snapshot is use only to create a ChunkSnapshotModel who need snapshot of the chunk and is neigbours. Need lot optimisation...

ghost: this line : Chunk chunk = world.getChunk(i, j, k);
I just want to get the instance of chunk, to to access to the data, i need the lock ?

ghost: I think it's better if a change on a chunk call the render to refresh the chunk, it's heavy to check each dirty state on all rendered chunk. So snapshot for render only maybe useless.

PR #2024, opened 2012-10-27
WorldRender use mesh and texture for block !

ghost:
Signed-off-by: Karang & L5D cholley.thibaut@gmail.com


src/main/java/org/spout/engine/mesh/ChunkMesh.java
```diff
@@ -152,16 +156,50 @@ private void generateBlockVertices(int x, int y, int z) {
 		}*/
 		
 		Vector3 model = new Vector3(x & Chunk.BLOCKS.MASK, y & Chunk.BLOCKS.MASK, z & Chunk.BLOCKS.MASK);
+
+		CubeMesh cubeMesh = WorldRenderer.blocksMesh.get(material.getDisplayName());
```
Afforess, line 160: Using the name (Strings) has to be slower than using the ids, no?

Karang, line 160: We didnt know the id :D they arent the same as minecraft

ghost, line 160: It's temporary because we don't know the spout id.
But it would be replaced by material.getModel().getMesh() when it's implemented in vanilla.

PR #2031, opened 2012-10-30
Chat colors implementation.

Karang:

### DIFF
--- a/src/main/java/org/spout/engine/SpoutClient.java
+++ b/src/main/java/org/spout/engine/SpoutClient.java
@@ -64,6 +64,7 @@ import org.spout.api.Client;
 import org.spout.api.FileSystem;
 import org.spout.api.Spout;
 import org.spout.api.audio.SoundManager;
+import org.spout.api.chat.style.ChatStyle;
 import org.spout.api.command.CommandRegistrationsFactory;
 import org.spout.api.command.CommandSource;
 import org.spout.api.command.annotated.AnnotatedCommandRegistrationFactory;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spout/engine/SpoutClient.java
```diff
@@ -503,7 +503,7 @@ public void initRenderer() {
 		Entity e = spoutyType.createEntity(new Point(super.getDefaultWorld(),0,0,0));
 		e.setSavable(false); // To prevent entity duplication
 		TextModelComponent tmc = e.add(TextModelComponent.class);
-		tmc.setText("Spouty");
+		tmc.setText(ChatStyle.BLUE+"Sp"+ChatStyle.WHITE+"ou"+ChatStyle.RED+"ty");
```
olliestanley, line 506: Shouldn't it take varargs for Object in setText then be

ChatStyle.BLUE, "Sp", ChatStyle.WHITE, "ou", etc

Because that's what is is everywhere else, unlike in Bukkit where ChatColors were just string codes really?

Same goes for the lines 546-551 down and line 242

Zidane:
@RoyAwesome @Karang 

setText should take a ChatArguments...

olliestanley:
I think there should also be Object... convenience method like everywhere else :P

src/main/java/org/spout/engine/SpoutClient.java
narrowtux, line 506: He didn't understand the concept :P

I will change the GUI system to use a ChatStyles list appropriately.

olliestanley, line 506: I was slightly wrong; the Object... is only a convenience method for ChatArguments (as Zidane said below) :P

PR #2038, opened 2012-11-01
ConcurrentSkipListSet for manage mesh priority

ghost:
Task to manage batch generation
Set TARGET_FPS to 180, we should make better than 60
Remove the default material in worldrenderer
Finalize and remove old batch more fast

### DIFF
--- a/src/main/java/org/spout/engine/world/SpoutRegion.java
+++ b/src/main/java/org/spout/engine/world/SpoutRegion.java
@@ -33,6 +33,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -178,13 +179,17 @@ public class SpoutRegion extends Region {
 		public int compare(final SpoutChunkSnapshotModel e1, final SpoutChunkSnapshotModel e2) {
 			int dist1 = e1.getX() + e1.getY() + e1.getZ();
 			int dist2 = e2.getX() + e2.getY() + e2.getZ();
+			
 			if(dist1 == dist2)
 				return 1;
+			
 			return dist1 - dist2;
 		}
 	};
 	
+	private final TInt21TripleObjectHashMap<SpoutChunkSnapshotModel> renderChunkQueued = new TInt21TripleObjectHashMap<SpoutChunkSnapshotModel>();
 	private final ConcurrentSkipListSet<SpoutChunkSnapshotModel> renderChunkQueue = new ConcurrentSkipListSet<SpoutChunkSnapshotModel>(ChunkOrdering);
+	
 	private final AtomicReference<SpoutRegion>[][][] neighbours;
 
 	@SuppressWarnings("unchecked")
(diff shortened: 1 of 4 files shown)


Spouty:
**[AUTO-GENERATED COMMENT]**

Raphfrk:
I don't think you need a skip list in SpoutRegion right?

I can see why it might be useful for the renderer.  However, the thread in spout region is just 
- read from queue
- process snapshot to mesh data
- place on another queue

This has nothing to do with ordering right ?

ghost:
The mesh generation is the part that take the most time, so we need to generate the meshs near the player in priority.
If i manage the priority in the next queue, it's too late.

src/main/java/org/spout/engine/world/SpoutRegion.java
```diff
@@ -167,7 +168,23 @@
...
+		//TODO : Compare with the chunk player distance
+		
+		@Override
+		public int compare(final SpoutChunkSnapshotModel e1, final SpoutChunkSnapshotModel e2) {
+			int dist1 = e1.getX() + e1.getY() + e1.getZ();
+			int dist2 = e2.getX() + e2.getY() + e2.getZ();
+			if(dist1 == dist2)
+				return 1;
+			return dist1 - dist2;
+		}
```
Raphfrk, line 187: The way the comparators work, they have to be constant while the object is in the queue.

You must decide the distance to the player when the snapshot is added to the queue and then never change it.

This should still work reasonably, as long as the player isn't moving to quickly.

There needs to be some way that snapshots for a given chunk are never reordered.

Raphfrk, line 187: I had a think about it and I think swapping to ConcurrentSkipListSet + ConcurrentHashMap (as a set) might be the way to go.

This would allow removal of snapshots if another dirty one arrived later.

The process would be something like
- Add to HashSet
- If there was an old value, remove it from SkipListSet
- Add to SkipList Set

The SkipListSet would use the distance value in the model to sort by

ghost, line 187: I agree

Raphfrk, line 187: What's nice is that it could prevent processing multiple dirty versions of the same chunk if they arrive together.

ghost, line 187: The HashSet should be sync too no ? When we remove a model to generate his mesh, should we remove in the hashset too ?

ghost, line 187: Maybe a TInt21TripleObjectHashMap<Model> to store with the position ? HashSet don't allow to get the old model with the newest.

Raphfrk, line 187: I think the triple int isn't required, since it is local to 1 Region.  It does need to be concurrent though.

But yeah, the .take() method would have to remove the object from the hashset.

ghost, line 187: What implementation of hashset should you use ?

Raphfrk, line 187: Triple byte probably.  I will have a look this evening at adding a class to SpoutAPI.

Maybe, I could call it "UniquePriorityBlockingQueue".  I could also add the synchronization stuff so it will block properly.

ghost, line 187: The synchronization should concern the hashset and the queue at same time, so a syncronized block ?

I writed that :

private void addToRenderQueue(SpoutChunkSnapshotModel model){
        synchronized (renderChunkQueue) {
            SpoutChunkSnapshotModel previous = renderChunkQueued.get(model.getX(), model.getY(), model.getZ());

```
        if(previous != null){
            if(previous.getTime() < model.getTime()){
                renderChunkQueued.put(model.getX(), model.getY(), model.getZ(), model);
                renderChunkQueue.remove(previous);
                renderChunkQueue.add(model);
            }
        }else{
            renderChunkQueued.put(model.getX(), model.getY(), model.getZ(), model);
            renderChunkQueue.add(model);
        }
    }
}

private SpoutChunkSnapshotModel removeFromRenderQueue(){
    synchronized (renderChunkQueue) {
        SpoutChunkSnapshotModel model = renderChunkQueue.pollFirst();
        if(model == null) return null;
        renderChunkQueued.remove(model.getX(), model.getY(), model.getZ());
        return model;
    }
}
```

Raphfrk, line 187: It is possible to do it with less direct synchronization.  However, maybe it is just overkill.

Also, you need to be careful when using the skip list.  The comparator must match the equals method.

If .equals(Object other) is based on the chunk position, then you shouldn't based the comparison on the distance.  Care is needed to get everything to match,

ghost, line 187: I remove the old model before add the newest, so the equals problem shouldn't happens.

And yeah, i agree, i mean the synchronization can be totaly useless on the renderChunkQueued.

PR #2046, opened 2012-11-06
Add a command to test block interactions

Karang:

### DIFF
--- a/src/main/java/org/spout/engine/command/TestCommands.java
+++ b/src/main/java/org/spout/engine/command/TestCommands.java
@@ -35,14 +35,19 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.List;
 
+import org.spout.api.Client;
 import org.spout.api.Spout;
 import org.spout.api.command.CommandContext;
 import org.spout.api.command.CommandSource;
 import org.spout.api.command.annotated.Command;
 import org.spout.api.command.annotated.CommandPermissions;
+import org.spout.api.component.components.HitBlockComponent;
 
+import org.spout.api.entity.Player;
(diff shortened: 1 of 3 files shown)


Spouty:
**[AUTO-GENERATED COMMENT]**

src/main/java/org/spout/engine/command/TestCommands.java
```diff
@@ -52,6 +57,19 @@ public TestCommands(SpoutEngine engine) {
...
+	public void debugBreak(CommandContext args, CommandSource source) {
+		if (Spout.getPlatform()!=Platform.CLIENT) {
+			
+		}
+		Player player = ((Client) Spout.getEngine()).getActivePlayer();
+		HitBlockComponent hbc = player.get(HitBlockComponent.class);
+		Block block = hbc.getTargetBlock();
+		source.sendMessage("Block to break: ", block.toString());
+		// TODO: A visual to test the code (maybe spawn an entity at the hit location)
+		// (I've tested it with coordinates and it seems to work)
```
Raphfrk: Why not just replace the block with air?

Jckf: Because that's not what he was testing?

Raphfrk: Yes it is.  

The command is debugBreak().  It even says "block to break".  If he also wants to test that drops are spawned, then that is separate, he still needs to clear the target block.

We also need a method which will return the face that is hit.

Karang:
I wasn't sure how to regenerate the chunk after clearing the block so I preferred to let it that way.

Raphfrk:
It should automatically regenerate.  In fact, that would be good to test.

PR #2055, opened 2012-11-09
Make a pool of BatchVertexRenderer

ghost:
Fix a equals method of aggregator


src/main/java/org/spout/engine/renderer/BatchVertexRenderer.java
```diff
@@ -42,8 +43,52 @@
...
+		List<Renderer> list = pool.get(renderMode);
+
+		if(list == null){
+			list = new LinkedList<Renderer>();
+			pool.put(renderMode, list);
+		}
+		
+		Client client = (Client) Spout.getEngine();
+
+		if (client.getRenderMode() == RenderMode.GL11) {
```
windy1, line 59: A switch statement probably would have been better here. :/

PR #2056, opened 2012-11-10
Use switch to get BatchVertexRenderer

ghost:
As wanted by @W1ndwaker


src/main/java/org/spout/engine/renderer/BatchVertexRenderer.java
```diff
@@ -56,27 +55,26 @@ public static void initPool(int renderMode, int number){
 		
 		Client client = (Client) Spout.getEngine();
 
-		if (client.getRenderMode() == RenderMode.GL11) {
+		switch (client.getRenderMode()) {
+		case GL11:
```
windy1: Cases should be tabbed from the switch statement.

ghost: Solved, my default indentation from eclipse don't make case tabbed.

PR #2079, opened 2012-11-17
Fix the PreCommandEvent being called before the command lookup. Fixes SPOUT-92

Snaipe:
SpoutAPI PR : https://github.com/SpoutDev/SpoutAPI/pull/424

Signed-off-by: Snaipe franklinmathieu@gmail.com


Snaipe:
Now, that's better.

PR #2097, opened 2012-11-23
Limit queue to the WorldRenderer

ghost:
Rewrite the renderer init

@Raphfrk if you cant take a look

I think we can improve more with leftView/enterView in a overload case (high speed of player), to avoid to send chunk to unload if it are never loaded.


RoyAwesome:
Can't be merged

PR #2103, opened 2012-11-24
Add Block.getRelative()

Karang:


DDoS:
This already exists, use Block#translate.

PR #2104, opened 2012-11-25
Send input to WidgetComponents

windy1:
SpoutAPI: https://github.com/SpoutDev/SpoutAPI/pull/442

Signed-off-by: Walker Crouse walkercrouse@hotmail.com


src/main/java/org/spout/engine/input/SpoutInputManager.java
```diff
@@ -0,0 +1,194 @@
...
+			throw new IllegalStateException("Cannot access ScreenStack in server mode.");
+		}
+
+		Set<WidgetComponent> components = new HashSet<WidgetComponent>();
+		Screen inputScreen = ((Client) engine).getScreenStack().getInputScreen();
+		if (inputScreen == null) {
+			return components;
+		}
+
+		for (Widget widget : inputScreen.getWidgets()) {
```
narrowtux, line 129: This can be reduced to the widget that is in focus since the other widgets shouldn't handle keyboard presses when they're not focussed.

PR #2113, opened 2012-11-30
Remove Aggregator, use one batch per chunk directly.

ghost:


src/main/java/org/spout/engine/scheduler/SpoutScheduler.java
```diff
@@ -211,8 +211,27 @@ public void run() {
...
+				if (timeError > 0) {
+					delay--;
+				} else if (timeError < 0) {
+					delay++;
+				}
+
+				if (delay > 0) {
+
+					c.updateRender(delay);
+
```
Raphfrk, line 234: Not sure this is right, shouldn't the elapsed time be taken into account at some point ?

ghost, line 234: Right, the second call of updateRender shouldn't be.

PR #2118, opened 2012-12-03
Plugin for exporting SAM with blender.

Karang:

### DIFF
--- a/tools/SAMExporter/Readme.txt
+++ b/tools/SAMExporter/Readme.txt
@@ -1,0 +1,19 @@
+#########################
+# Spout Animation Model #
+#########################
+
+Author: Karang
+Version: v1.0
+
+/!\ This is still in developpement, I only released it because Roy asked to :D
+
+HOW TO INSTALL:
+Drag the io_anim_sam into your blender's addon folder.
+Restart blender.
+Go into File>User preferences
+Go in the tab Addon
+Find the plugin then tick the box to enable it.
+
+HOW TO USE:
+Select you mesh (not the armature)
+Then go into File>Export>Spout Animation Model (SAM)
--- a/tools/SAMExporter/io_anim_sam/__init__.py
+++ b/tools/SAMExporter/io_anim_sam/__init__.py
@@ -1,0 +1,100 @@
+import bpy
+from bpy_extras.io_utils import ExportHelper
+
+bl_info = {
+    "name": "Spout Animation Model",
+    "author": "Karang",
+    "blender": (2,6,2),
+    "version": (0,0,1),
+    "location": "File > Import-Export",
+    "description": "Export bones animation for spout engine.",
+    "category": "Import-Export"
+}
+
+class ExportSAM(bpy.types.Operator, ExportHelper):
+    bl_idname = "export_anim.sam";
+    bl_label = "Spout Animation Model";
+    bl_options = {'PRESET'};
+    
+    filename_ext = ".sam";
+
+    def saveSAM(self, context, path):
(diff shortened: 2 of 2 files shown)


src/main/resources/tools/io_anim_sam/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    bl_label        = "Spout Animation Model";
+    bl_options      = {'PRESET'};
+    
+    filename_ext    = ".sam";
+
+    def saveSAM(self, context, path):
+        f = open(path, "w", encoding="utf8", newline="\n");
+		
+		f.write("# File exported by Blender's Spout Animation Model plugin");
+		f.write("# Tool made by: Karang");
```
Karang: hum someone needs to put indent on theses lines, last edit in eclipse wasn't such a good idea :D

src/main/resources/tools/io_anim_sam/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+
+class ExportSAM(bpy.types.Operator, ExportHelper):
+    bl_idname       = "export_anim.sam";
+    bl_label        = "Spout Animation Model";
+    bl_options      = {'PRESET'};
+    
+    filename_ext    = ".sam";
+
+    def saveSAM(self, context, path):
+        f = open(path, "w", encoding="utf8", newline="\n");
```
zml2008: use with open(path, "wb", encoding="utf8", newline="\n") as f:
and indent everything that uses the file
and pass the file to all the methods that need it

src/main/resources/tools/io_anim_sam/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    "name":         "Spout Animation Model",
+    "author":       "Karang",
+    "blender":      (2,6,2),
+    "version":      (0,0,1),
+    "location":     "File > Import-Export",
+    "description":  "Export bones animation for spout engine.",
+    "category":     "Import-Export"
+}
+        
+import bpy
```
zml2008: imports go first: http://www.python.org/dev/peps/pep-0008/#imports

src/main/resources/tools/io_anim_sam/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+            for group in v.groups:
+                vgroups[obj.vertex_groups[group.group].name].append(v.index);
+
+        children = {None: []};
+        for bone in armature.bones:
+            children[bone.name] = [];
+        for bone in armature.bones:
+            children[getattr(bone.parent, "name", None)].append(bone.name);
+        
+        def writeBones(bone, indent):
```
zml2008: move this method out, add file param

zml2008:
This probably shouldn't be in src/main/resources - that's for stuff in the jar. Stick it in a tools or etc folder in the repo root. Since your module is only one file, it shouldn't be a package -- stick the io_anim_sam.py and README.txt in a subfolder - this isn't the only tool.

src/main/resources/tools/io_anim_sam/__init__.py
Karang: For my defense I would say it has been 4 years since I last did python. But you're right ;)

src/main/resources/tools/io_anim_sam/__init__.py
Karang: Seems good, Thanks for the advices

Karang:
I cleaned the PR a bit

PR #2142, opened 2012-12-11
Add VertexEffect

DDoS:
Adds the ability to insert suplemental arbitrary vertex data taken from the world.

This data must be a float or float structure and interpretable by OpenGL (as matrices, vectors, etc.). This data is then passed to the vertex shader as an "in" variable.

The data is taken from one or more VertexEffect during addition of the vertex data to the buffers. Each VertexEffect gets its own buffer.

Each VertexEffect corresponds to one suplemental "in" variable for the vertex shader.

VertexEffects are added to RenderMaterials by plugins.

Related PR:
https://github.com/SpoutDev/SpoutAPI/pull/463

Example of usage:
Vanilla: https://github.com/VanillaDev/Vanilla/pull/408

Signed-off-by: Aleksi Sapon QcTechs@gmail.com

### DIFF
--- a/src/main/java/org/spout/engine/mesh/ChunkMesh.java
+++ b/src/main/java/org/spout/engine/mesh/ChunkMesh.java
@@ -26,10 +26,14 @@
  */
 package org.spout.engine.mesh;
 
+import gnu.trove.iterator.TIntObjectIterator;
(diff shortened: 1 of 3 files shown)


DDoS:
Will redo with a buffer effect instead.

PR #2150, opened 2012-12-24
Added getOnlinePlayers() to SpoutEngine

thehutch:
I'm adding this for a broadcastMessage()

### DIFF
--- a/src/main/java/org/spout/engine/SpoutEngine.java
+++ b/src/main/java/org/spout/engine/SpoutEngine.java
@@ -703,10 +703,10 @@ public abstract class SpoutEngine extends AsyncManager implements Engine {
 	}
 
 	@Override
-	public List<String> getAllPlayers() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/engine/SpoutEngine.java
```diff
@@ -701,10 +701,19 @@ public Entity getEntity(UUID uid) {
 		}
 		return null;
 	}
+	
+	@Override
+	public List<Player> getOnlinePlayers() {
+		List<Player> players = new ArrayList<Player>();
+		for(String name : onlinePlayers.keySet()) {
+			players.add(onlinePlayers.get(name));
```
Afforess, line 709: This is a really slow way of iterating a map. You can do this instead of all that code:

`return new ArrayList<Player>(onlinePlayers.values());`

zml2008:
There's already a broadcastMessage(). What are you trying to add?

PR #2151, opened 2012-12-27
hideEntity and showEntity

kitskub:

Vanilla: https://github.com/VanillaDev/Vanilla/pull/451
SpoutAPI: https://github.com/SpoutDev/SpoutAPI/pull/481

### DIFF
--- a/src/main/java/org/spout/engine/entity/SpoutPlayer.java
+++ b/src/main/java/org/spout/engine/entity/SpoutPlayer.java
@@ -27,6 +27,9 @@
 package org.spout.engine.entity;
 
 import java.net.InetAddress;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spout/engine/entity/SpoutPlayer.java
```diff
@@ -378,4 +386,25 @@ public void preSnapshotRun() {
...
+			Spout.getEventManager().callEvent(new EntityShowEvent(entity, this));
+		} else {
+			hiddenEntities.add(entity);
+			Spout.getEventManager().callEvent(new EntityHideEvent(entity, this));
+		}
+	}
+
+	@Override
+	public List<Entity> getInvisibleEntities() {
+		return new ArrayList<Entity>(hiddenEntities);
```
Zidane, line 403: This is not thread-safe

kitskub, line 403: new instance each time. Would you rather have Collections.immutableList?

PR #2156, opened 2012-12-31
Fix broken links in README

rocode:
Fixed Vanilla and BukkitDev links in README file to point to their respective new addresses.


lukespragg:
This is fixed locally. Please do not submit PRs for the readme or pom.xml.

PR #2158, opened 2012-12-31
resolves Vanilla-227

Sleaker:
Not sure why there is an index here with arguments stuff, it looks sloppy, but this should fix the incorrect name showing as offline.

Signed-off-by: Sleaker sleaker@gmail.com


lukespragg:
This did not fix the issue by the way.

Sleaker:
Fixed it for me....  Looks like there might be more issues though, there was another reference of target down below.  Why are we using index btw? Seems confusing.  Also there's a args % 2 and an args > 2 does anyone realize that this is possible to trigger both?

I think I might just rewrite this command completely.

PR #2160, opened 2013-01-02
Handle TP Command a bit better

Sleaker:
See: https://spoutdev.atlassian.net/browse/SPOUT-162

Signed-off-by: Sleaker sleaker@gmail.com


src/main/java/org/spout/engine/command/AdministrationCommands.java
```diff
@@ -435,58 +435,54 @@ public void worldInfo(CommandContext args, CommandSource source) throws CommandE
...
 	public void tp(CommandContext args, CommandSource source) throws CommandException {
-		int index = 0;
 		Player player;
-
-		if (args.length() % 2 == 0) {
-			if (Spout.getEngine() instanceof SpoutClient) {
-				throw new CommandException("You cannot search for players unless you are in server mode.");
+		Player target = null;
+		Point point;
+		if (args.length() % 2 == 0 && (Spout.getEngine() instanceof SpoutClient)) {
```
zml2008: Use the engine field that's already passed to the commands.

Sleaker: why was the previous implementation not doing that? haha. I'll fix it.

Sleaker:
alright @zml2008 - that look better, also fixed pulling the incorrect target.

PR #2163, opened 2013-01-19
Seperated out the username a bit more when joining

Grinch:


Grinch:
Updated for when a player leaves as well.

PR #2193, opened 2013-03-13
Allow everybody to chat.

greatman:
This goes with https://github.com/SpoutDev/SpoutAPI/pull/542

Signed-off-by: William Lebel lebelwilliam@gmail.com


zml2008:
um, you use DefaultPermissions to do this. The default value for permissions is false...

PR #2195, opened 2013-05-07
Add onPlayerEnter/onPlayerExit in world components

Snaipe:
Goes with https://github.com/SpoutDev/SpoutAPI/pull/549
signed-off-by: Snaipe franklinmathieu@gmail.com

### DIFF
--- a/src/main/java/org/spout/engine/listener/SpoutServerListener.java
+++ b/src/main/java/org/spout/engine/listener/SpoutServerListener.java
@@ -29,6 +29,8 @@ package org.spout.engine.listener;
 import java.net.InetAddress;
 
 import org.spout.api.chat.style.ChatStyle;
+import org.spout.api.component.Component;
+import org.spout.api.component.type.WorldComponent;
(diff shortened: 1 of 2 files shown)


Afforess:
I agree with the idea of this pull request, but unfortunately the implementation is a bit more complicated. I suggested events in the API pull request. These events would be fired in the location your code is currently, but as a delayed event (add/remove player may not be in the main api tickstage, so exposing this event/callback is bad). Then, in the Spout Engine internal event listener, listen to both events, and iterate the world components to call the callback methods. This ensures that the callbacks are executed in the main tickstage.

Also, the callbacks need try/catch to ensure any exceptions from plugins do not prevent other components from receiving the callback, nor break the player add/removal process.

dredhorse:
should this event be cancellable btw?

Snaipe:
No, a player enters a world by multiple means, be it at login or when it
teleports. Denying access should be made in their respective events

dredhorse:
true... forget my comment than

PR #2200, opened 2013-05-18
Only restrict getChunk to certain stages if actions are required

kitskub:


Afforess:
No. While I understand the reasoning, this could mask potentially dangerous/illegal API calls, when they should fail fast and force developers to rethink how they are using getChunk.

PR #2201, opened 2013-05-27
Removed unnecessary exception for no world plugins

thehutch:
Signed-off-by:thehutch

### DIFF
--- a/src/main/java/org/spout/engine/SpoutEngine.java
+++ b/src/main/java/org/spout/engine/SpoutEngine.java
@@ -252,8 +252,9 @@ public abstract class SpoutEngine implements AsyncManager, Engine {
 
 		if (checkWorlds) {
 			//At least one plugin should have registered atleast one world
-			if (loadedWorlds.getLive().size() == 0) {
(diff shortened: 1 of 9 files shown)


src/main/java/org/spout/engine/SpoutEngine.java
```diff
@@ -853,6 +854,7 @@ public CompletionManager getCompletionManager() {
 		return completions;
 	}
```
windy1, line 855: I don't think this is any better to be honest, this is the kind of place where an IllegalStateException is appropriate.

thehutch, line 855: Except having a stacktrace is completely pointless, that's why I removed it. Yes, it's in a illegal state but a stacktrace isn't necessary just a simple log message

kitskub, line 855: Agreed.

dredhorse, line 855: seconded

PR #2204, opened 2013-06-02
Format

thehutch:
Thought these branches wouldn't merge :(


macalinao:
Don't forget to rebase James :smiling_imp:

windy1:
Correct me if I'm wrong but, I get the impression that a good portion of this was automated. This can be useful sometimes, but I've learned from experiences that these analyzers typically reduce readability.

PR #2211, opened 2013-06-09
Numerous fixes and such

kitskub:
Fix tp command
Fix the variable naming in LoginMessage to reflect what it actually is
Changed NOTSPAWNEDID to be -999 because -1 was/is being used as a player id (this needs polishing)
Refactored start(boolean) in SpoutEngine to just start()
Fix my session pulsing error from last PR

### DIFF
--- a/src/main/java/org/spout/engine/SpoutClient.java
+++ b/src/main/java/org/spout/engine/SpoutClient.java
@@ -96,6 +96,7 @@ public class SpoutClient extends SpoutEngine implements Client {
 	private final ClientBootstrap bootstrap = new ClientBootstrap();
 	private final FileSystem filesystem = new ClientFileSystem();
 	private final AtomicReference<SpoutClientPlayer> player = new AtomicReference<SpoutClientPlayer>();
+	private final SessionTask sessionTask = new SessionTask();
 	// Handle stopping
 	private volatile boolean rendering = true;
 	private boolean ccoverride = false;
@@ -145,11 +146,6 @@ public class SpoutClient extends SpoutEngine implements Client {
 
 	@Override
 	public void start() {
-		start(true);
-	}
-
-	@Override
-	public void start(boolean checkWorlds) {
 		if (!connnect()) {
 			return;
 		}
@@ -159,7 +155,7 @@ public class SpoutClient extends SpoutEngine implements Client {
 
 		// Completely blank world
 		worldChanged("NullWorld", UUID.randomUUID(), new DatatableComponent().serialize());
-		super.start(checkWorlds);
+		super.start();
 
 		getEventManager().registerEvents(new SpoutClientListener(this), this);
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/spout/engine/protocol/builtin/message/LoginMessage.java
```diff
@@ -33,26 +33,26 @@
 
 public class LoginMessage extends SpoutMessage {
 	private final String playerName;
-	private final int protoVersion;
```
zml2008, line 36: It's actually used as both, protocol version C->S and player id S->C

windy1:
Why does this appear to be amended onto @Afforess's commit?

src/main/java/org/spout/engine/protocol/builtin/message/LoginMessage.java
kitskub, line 36: Hmm. You're right: the client is putting the protocol version in the message. But the server isn't using it.

zml2008, line 36: I don't think I ever got around to implementing protocol version checking, which is probably why.

kitskub, line 36: Well, I went ahead and named it appropriately (extraInt)

kitskub:
@W1ndwaker I think I accidently amended the commit. Fixed.

src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -73,14 +73,14 @@
 import org.spout.engine.world.SpoutRegion;
 
 public class SpoutEntity extends BaseComponentOwner implements Entity, Snapshotable {
-	public static final int NOTSPAWNEDID = -1;
+	public static final int NOTSPAWNEDID = -999;
```
windy1: If we're allowing negative Entity IDs then why wouldn't we use `Integer.MIN_VALUE`

src/main/java/org/spout/engine/protocol/builtin/message/LoginMessage.java
windy1, line 36: "extraInt" is incredibly ambiguous.

kitskub, line 36: Well, I COULD make it playerIdOrProtocolVersion?

src/main/java/org/spout/engine/SpoutClient.java
```diff
@@ -226,6 +221,16 @@ private boolean connnect() {
 	}
 
 	@Override
+	protected Runnable getSessionTask() {
+		return new Runnable() {
```
Afforess: One minor suggestion. Make this a private final variable, and create the runnable once. Then return the variable instead. The outcome is the same, with the advantage of not creating an object every tick.

src/main/java/org/spout/engine/SpoutServer.java
```diff
@@ -320,17 +315,14 @@ public SessionRegistry getSessionRegistry() {
...
-			this.registry = registry;
-		}
-
-		@Override
-		public void run() {
-			registry.pulse();
-		}
+	@Override
+	protected Runnable getSessionTask() {
+		return new Runnable() {
```
Afforess: Same as earlier. Also, I had the code in a private class on purpose. When exceptions are thrown, it's much easier to pick out the cause with named inner classes versus the anonymous inner class names that are autogenerated.

kitskub:
Fixed. In my defense, I thought when the task was scheduled, only one instance was created.

PR #2212, opened 2013-06-10
Add protocol tests

kitskub:
And more

ChunkDataCodec is fixed. 
Client starts without error.
resend command for Server


Afforess:
Unit Tests make me happy.

PR #2217, opened 2013-06-15
Feedback and PR of Chunk Mesh fix

kitskub:
This is my attempt at readding the chunk meshing. It currently DOESN'T work. So, please give feedback where you see it broken. @Zidane @RoyAwesome @Afforess 

DISCLAIMER: I am not fluent in rendering.

NOTE: this is also rebased on client-connection at the request of @Zidane


Zidane:
This is actually not right at all but thats fine.

When you rebased meshes into client-connection you actually didn't remove the code and leave it gone. The MeshGeneratorTask is completely wrong as well as it shouldn't be a queue of chunks.

It should just send off async tasks and check the sequences before generating a mesh. I'll fix things up right for ya.

PR #2218, opened 2013-06-15
merge client-connection on master

kitskub:
At the request of @Zidane and @Afforess


kitskub:
@Afforess these are updated

PR #2224, opened 2013-07-12
Replace all instances of <p/> with <p> to follow Oracle standards.

Grinch:
http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format


Afforess:
This is still just as bad. < p > is the _opening tag_ of a paragraph. You also need to add _closing tags_ (< /p >) at the end of the paragraphs. Writing invalid html is not cool.

Grinch:
Looking at the examples in that page they never once use the closing tag at all. Even in cases where there are multiple paragraphs.

DDoS:
Grinch is right.

Afforess:
Stackoverflow Agrees: http://stackoverflow.com/questions/5077382/javadoc-line-breaks-without-html-tags

It seems horrible. Such is life.

macalinao:
HTML != XHTML

PR #2228, opened 2013-07-17
Add version command

Grinch:
- Adds /version (with /ver as an alias)
- Adds 'spout.command.version' permission to the 'version' command.
- Adds 'spout.command.batch' permission to the 'batch' command.
- Changes the server version output displayed on startup.

### DIFF
--- a/src/main/java/org/spout/engine/command/CommonCommands.java
+++ b/src/main/java/org/spout/engine/command/CommonCommands.java
@@ -27,6 +27,7 @@
 package org.spout.engine.command;
 
 import org.spout.api.Server;
+import org.spout.api.Spout;
 import org.spout.api.command.CommandArguments;
 import org.spout.api.command.CommandBatch;
 import org.spout.api.command.CommandSource;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spout/engine/command/CommonCommands.java
```diff
@@ -258,13 +260,20 @@ public void tp(CommandSource source, CommandArguments args) throws CommandExcept
 
 	@CommandDescription(aliases = "validate_movement", desc = "Toggle the validating of movement.")
 	@Filter(PlayerFilter.class)
-	public void validateInput(Player player, CommandArguments args) throws CommandException {
```
zml2008, line 261: This should work because PlayerFilter.

Grinch, line 261: Spout threw an error registering it at startup without having 'CommandSource'.

zml2008, line 261: Fixed

Grinch:
Updated to reflect @zml2008's recent fix in SpoutAPI for validateInput.

macalinao:
Please also add /version [plugin]

kitskub:
I agree with @simplyianm . Can you add that?

Grinch:
@kitskub @simplyianm Perhaps in a future PR but as of right now I really have little intention to add such a feature. The goal was to stop relying on Vanilla's /version command and give Spout its own finally.

kitskub:
Ok, sounds fine.

PR #2230, opened 2013-07-28
Removed empty constructor as developers should use VectorX.ZERO for zero'ed vector

thehutch:
Signed-off-by:thehutch th3hutch@yahoo.co.uk

### DIFF
--- a/engine/src/main/java/org/spout/engine/filesystem/resource/loader/CubeMeshLoader.java
+++ b/engine/src/main/java/org/spout/engine/filesystem/resource/loader/CubeMeshLoader.java
@@ -41,87 +41,83 @@ public class CubeMeshLoader extends ResourceLoader {
 	}
 
 	private static OrientedMesh loadObj(InputStream stream) {
-		Scanner scan = new Scanner(stream);
-
-		ArrayList<Vector2[]> textures = new ArrayList<Vector2[]>();
-		ArrayList<Vector2> uvs = new ArrayList<Vector2>();
-		Vector2 scale = Vector2.ONE;
-		Vector2 size = Vector2.ONE;
-		Vector2 sizeScaled = Vector2.ONE;
-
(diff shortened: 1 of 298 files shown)


math/src/test/java/org/spout/math/test/vector/Vector2Test.java
```diff
@@ -416,7 +411,7 @@ public void testCloning() {
 	@Test
 	public void testCreateDirectionFromRandom() {
 		Random random = new Random();
-		for (int i = 0; i < 10; i++) {
+		for (int i = 0 ; i < 10 ; i++) {
```
kitskub, line 414: I'm pretty sure Spout formatting is for (int i = 0; i < 10; i++) {

thehutch, line 414: Didn't mean to change this :P I didn't know I had until I made the PR

kitskub, line 414: Probably should be fixed.

engine/src/main/java/org/spout/engine/filesystem/resource/loader/CubeMeshLoader.java
```diff
@@ -41,87 +41,83 @@ public CubeMeshLoader() {
 	}
 
 	private static OrientedMesh loadObj(InputStream stream) {
-		Scanner scan = new Scanner(stream);
```
kitskub, line 44: Is this breaking or fixing?

thehutch, line 44: It's a convert to try-with-resources

kitskub, line 44: Okay

PR #2231, opened 2013-07-28
Deleted empty constructors in Vector classes as developers should use VectoX.ZERO  instead

thehutch:
- changed code to conform with JDK 7 standards
- try-with-resources
- use of diamond operator
- use of compound operators
- removed redundant code
- added @Override annotation
- more...
  Signed-off-by:thehutch th3hutch@yahoo.co.uk


kitskub:
By the approval of @Afforess and the goodness of this commit, it shall be merged.

PR #2233, opened 2013-07-30
Removed Session from SpoutPlayer into the PlayerNetworkComponent

thehutch:


engine/src/main/java/org/spout/engine/SpoutServer.java
```diff
@@ -734,7 +731,8 @@ public Player addPlayer(String playerName, SpoutServerSession<?> session, int vi
 		boolean created = false;
 		if (player == null) {
 			getLogger().info("First login for " + playerName + ", creating new player data");
-			player = new SpoutPlayer(this, playerName, getDefaultWorld().getSpawnPoint(), viewDistance);
+			player = new SpoutPlayer(this, playerName, getDefaultWorld().getSpawnPoint(), session);
+			player.getNetwork().setSyncDistance(viewDistance);
```
kitskub: The add(NetworkComponent.class) in SpoutEntity is going to be  removed; instead, it will be the job of initialize..Session to do that. Therefore, this will NPE.

kitskub:
Besides that one comment, this commit looks sound and mergable.

@Zidane?

engine/src/main/java/org/spout/engine/entity/SpoutPlayerSnapshot.java
```diff
@@ -71,6 +71,6 @@ public String getName() {
 	@SuppressWarnings ("unchecked")
 	@Override
 	public SpoutPlayer toEntity(SpoutEngine engine) {
-		return new SpoutPlayer(engine, name, getTransform(), getViewDistance(), getUID(), false, getDataMap(), getComponents().toArray(new Class[0]));
+		return new SpoutPlayer(engine, getName(), getTransform(), getUID(), false, getDataMap(), getComponents().toArray(new Class[0]));
```
kitskub, line 74: This method needs to be rethought. @Zidane do you agree?

thehutch, line 74: A SpoutSession needs to be put in there for it to work

Afforess, line 74: Not sure how you are going to pull this off. You need to keep it compatible so that plugins, if desired, can override the saving/loading mechanism and use their own (I have an example plugin that uses a database for player data, for example). If you break this piece, you break the entire point of having the feature.

kitskub, line 74: @Afforess I will take a look at this tomorrow.

engine/src/main/java/org/spout/engine/SpoutServer.java
Zidane: Not exactly, by default all entities should have the basic NetworkComponent. If I do getNetwork off of an entity, I should be confident in knowing it will never NPE. Remember that NetworkComponent serves two different purposes. Of course Player is different.

Player should have the Player version attached in initializeSession. That part is correct

@thehutch Change the naming from viewDistance to syncDistance.

api/src/main/java/org/spout/api/entity/Entity.java
```diff
@@ -155,4 +152,18 @@
 	 */
 	@Override
 	public ManagedMap getData();
+
+	/**
+	 * Teleports the entity to the given point
+	 *
+	 * @param transform Teleport location
+	 */
+	public void teleport(Point point);
```
Zidane: These methods make no sense anymore and I removed it for a good reason.

Spout has the ability to set a transform without sync, use it.

engine/src/main/java/org/spout/engine/entity/EntityManager.java
```diff
@@ -237,9 +237,10 @@ private void syncEntity(Entity ent, Set<? extends Entity> observers, boolean for
 				continue;
 			}
 			//Grab the NetworkSynchronizer of the player
-			ServerNetworkSynchronizer network = (ServerNetworkSynchronizer) player.getNetworkSynchronizer();
+			PlayerNetworkComponent network = player.getNetwork();
+			//ServerNetworkSynchronizer network = (ServerNetworkSynchronizer) player.getNetworkSynchronizer();
 			//Grab player's view distance
-			int view = player.getViewDistance();
+			int view = network.getSyncDistance();
```
Zidane: Again naming

engine/src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -285,7 +281,7 @@ protected void removeObserver() {
 
 	protected void updateObserver() {
 		List<Vector3> ungenerated = new ArrayList<>();
-		final int viewDistance = getViewDistance() >> Chunk.BLOCKS.BITS;
+		final int viewDistance = network.getSyncDistance();
```
Zidane: Again naming

engine/src/main/java/org/spout/engine/entity/SpoutEntitySnapshot.java
```diff
@@ -67,8 +68,9 @@ public SpoutEntitySnapshot(Entity e) {
 		this.location = e.getPhysics().getTransform();
 		this.worldName = e.getWorld().getName();
 		this.worldId = e.getWorld().getUID();
-		this.viewDistance = e.getViewDistance();
-		this.observer = e.isObserver();
+		final NetworkComponent network = e.get(NetworkComponent.class);
+		this.viewDistance = network.getSyncDistance();
```
Zidane: more naming

engine/src/main/java/org/spout/engine/entity/SpoutPlayer.java
```diff
@@ -424,4 +403,9 @@ public boolean isInvisible(Entity entity) {
 	public PlayerSnapshot snapshot() {
 		return new SpoutPlayerSnapshot(this);
 	}
+
+	@Override
+	public PlayerNetworkComponent getNetwork() {
+		return getExact(PlayerNetworkComponent.class);
```
Zidane: This kills the ability for Vanilla to extend PlayerNetworkComponent.

thehutch: Can't you override this method and have VanillaPlayer?

engine/src/main/java/org/spout/engine/SpoutServer.java
kitskub: Oh yes. I keep forgetting that you added in observer code.

engine/src/main/java/org/spout/engine/entity/SpoutPlayer.java
kitskub: No....just do get(NetworkComponent.class)

thehutch: I have to use get(PlayerNetworkComponent.class); Or I have to cast it.

Zidane: Like Entity, Player is an unextendable class outside of the Engine (I mean you could but you can't inject it into the engine without it dying from ClassCastExceptions).

Just do a get(PlayerNetworkComponent.class) there and cast it. We'll guarantee that Players must have an instance of PlayerNetworkComponent to prevent possible exceptions.

Vanilla is the one who would do getExact (as I don't want people interfering with Vanilla's packets).

engine/src/main/java/org/spout/engine/entity/SpoutPlayerSnapshot.java
Zidane, line 74: Fixing this method largely depends on where it is called.

If the presumption is "Snapshots are mostly used for offline players and saving content" then we can safely add a protected constructor to SpoutPlayer ad do a null session. Javadocs will be needed to explain the only potential way for getSession to not return a valid reference.

I haven't looked at the code but I am guessing the engine does not use a snapshot to load an entity in. In that case, we are fine.

api/src/main/java/org/spout/api/component/entity/PlayerNetworkComponent.java
```diff
@@ -43,7 +44,8 @@
  */
 public abstract class PlayerNetworkComponent extends NetworkComponent {
 	private static final SyncedStringMap protocolMap = SyncedStringMap.create(null, new MemoryStore<Integer>(), 0, 256, "componentProtocols");
-	private AtomicReference<Session> session = new AtomicReference<>(null);
+	private final AtomicReference<Session> sessionLive = new AtomicReference<>(null);
```
kitskub: Remove sessionLive. There's not reason for it. Just directly set session.

engine/src/main/java/org/spout/engine/SpoutServer.java
```diff
@@ -729,12 +726,13 @@ public Entity getEntity(UUID uid) {
...
 	// Players should use weak map?
-	public Player addPlayer(String playerName, SpoutServerSession<?> session, int viewDistance) {
+	public Player addPlayer(String playerName, SpoutServerSession<?> session, int syncDistance) {
 		SpoutPlayer player = PlayerFiles.loadPlayerData(playerName);
 		boolean created = false;
 		if (player == null) {
 			getLogger().info("First login for " + playerName + ", creating new player data");
-			player = new SpoutPlayer(this, playerName, getDefaultWorld().getSpawnPoint(), viewDistance);
+			player = new SpoutPlayer(this, playerName, getDefaultWorld().getSpawnPoint());
+			player.getNetwork().setSyncDistance(syncDistance);
```
kitskub: Move this below initializeServerSession

engine/src/main/java/org/spout/engine/command/CommonCommands.java
```diff
@@ -246,7 +246,7 @@ public void tp(CommandSource source, CommandArguments args) throws CommandExcept
 		args.assertCompletelyParsed();
 
 		point.getWorld().getChunkFromBlock(point);
-		player.teleport(point);
+		//TODO: player.teleport(point);
```
kitskub: player.getPhysics().setPosition(point)

engine/src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -129,7 +127,11 @@ public SpoutEntity(Engine engine, Transform transform, UUID uid, boolean load, b
 		observer.set(NOT_OBSERVING);
 		physics = (SpoutPhysicsComponent) add(PhysicsComponent.class);
 
-		network = add(NetworkComponent.class);
+		if (!(this instanceof Player)) {
+			network = add(NetworkComponent.class);
+		} else {
+			network = null;
```
kitskub: This needs to be rethought. initializeSession needs to be able to set the network. Or something comparable.

engine/src/main/java/org/spout/engine/entity/SpoutPlayer.java
```diff
@@ -385,7 +352,7 @@ public void finalizeRun() {
 				((SpoutServer) getEngine()).removePlayer(this);
 			}
 			// TODO stop client?
-			sessionLive.set(null);
+			getNetwork().setSessionLive(null);
```
kitskub: Remove setSessionLive

engine/src/main/java/org/spout/engine/entity/SpoutPlayerSnapshot.java
kitskub, line 74: This discussion is now unneeded.

engine/src/main/java/org/spout/engine/SpoutServer.java
```diff
@@ -762,10 +759,12 @@ public Player addPlayer(String playerName, SpoutServerSession<?> session, int vi
 
 		// Initialize the session
 		session.getProtocol().initializeServerSession(session);
+		//Set the player's sync distance
+		player.getNetwork().setSyncDistance(syncDistance);
```
kitskub: Why would you put this BEFORE the null check?

engine/src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -403,13 +400,21 @@ public boolean isSavable() {
...
 	public NetworkComponent getNetwork() {
 		return network;
 	}
 
 	@Override
-	public PhysicsComponent getPhysics() {
-		return physics;
+	public <T extends Component> T add(Class<T> type) {
+		if (network == null && NetworkComponent.class.isAssignableFrom(type)) {
+			this.network = super.add(NetworkComponent.class);
```
kitskub: This won't work for SpoutEntitys because add(..) doesn't REPLACE an existing component. You need to have custom logic here.

engine/src/main/java/org/spout/engine/protocol/NetworkSendThread.java
```diff
@@ -41,10 +41,10 @@
 	private final static long minimumLatency = SpoutConfiguration.SEND_LATENCY.getLong();
 	private final static long spikeLatency = SpoutConfiguration.SEND_SPIKE_LATENCY.getLong();
 	private final static float spikeChance = SpoutConfiguration.SEND_SPIKE_CHANCE.getFloat() / 10.0F;
-	private final int poolIndex;
-	private final AtomicReference<ChannelQueueThread[]> channelQueues = new AtomicReference<>();
 	private final AtomicReference<ChannelQueueThread[]> interruptedQueues = new AtomicReference<>();
-
+	private final AtomicReference<ChannelQueueThread[]> channelQueues = new AtomicReference<>();
```
kitskub, line 45: Why this diff?

engine/src/main/java/org/spout/engine/SpoutServer.java
thehutch: You just said to put it after the init

engine/src/main/java/org/spout/engine/protocol/SpoutServerSession.java
```diff
@@ -104,8 +103,8 @@ public void dispose() {
 	}
 
 	public void dispose(PlayerLeaveEvent leaveEvent, boolean stop) {
-		SpoutPlayer player;
-		if ((player = this.player.getAndSet(null)) != null) {
+		SpoutPlayer player = getPlayer();
+		if (player != null) {
```
kitskub, line 107: This is not the same logic.

engine/src/main/java/org/spout/engine/SpoutServer.java
kitskub: Well, yes: reasonably.

engine/src/main/java/org/spout/engine/protocol/SpoutServerSession.java
thehutch, line 107: I know but the field should be final and not protected, also if it's being disposed the fields will/should be deleted by the GC so no need to set null.

engine/src/main/java/org/spout/engine/protocol/NetworkSendThread.java
thehutch, line 45: preference of field code length :P

engine/src/main/java/org/spout/engine/protocol/SpoutServerSession.java
kitskub, line 107: Setting it to null will prevent the Session from being used after it is disposed.

thehutch, line 107: The session should be made null right after it is disposed anyway?

kitskub, line 107: It should be, yes.

engine/src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -403,13 +399,24 @@ public boolean isSavable() {
...
 		return network;
 	}
 
 	@Override
-	public PhysicsComponent getPhysics() {
-		return physics;
+	public <T extends Component> T add(Class<T> type) {
+		if (network == null && NetworkComponent.class.isAssignableFrom(type)) {
+			//Detach old NetworkComponent
+			detach(NetworkComponent.class);
```
kitskub: will be super.detach

engine/src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -403,13 +399,24 @@ public boolean isSavable() {
...
+			//Attach new one
+			this.network = super.add(NetworkComponent.class);
+		}
+		return super.add(type);
```
kitskub: return network;

engine/src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -403,15 +401,26 @@ public boolean isSavable() {
...
+		if (network == null && NetworkComponent.class.isAssignableFrom(type)) {
+			//Detach old NetworkComponent
+			super.detach(NetworkComponent.class);
+			//Attach new one
+			this.network = super.add(NetworkComponent.class);
+		}
+		return super.add(type);
 	}
-
+
```
kitskub: Whitespace

engine/src/main/java/org/spout/engine/entity/SpoutEntity.java
```diff
@@ -403,13 +401,25 @@ public boolean isSavable() {
...
+	@Override
 	public NetworkComponent getNetwork() {
 		return network;
 	}
 
 	@Override
-	public PhysicsComponent getPhysics() {
-		return physics;
+	public <T extends Component> T add(Class<T> type) {
+		if (network == null && NetworkComponent.class.isAssignableFrom(type)) {
```
kitskub: Remove network == null

We want to be able to reassign at will.

kitskub:
Looks fine to me (finally). @Zidane I'm pulling

PR #2235, opened 2013-08-01
Add debugging of sent/received protocol codes. May be overly verbose.

Sleaker:
Signed-off-by: Nick Minkler sleaker@gmail.com

### DIFF
--- a/api/src/main/java/org/spout/api/protocol/CommonDecoder.java
+++ b/api/src/main/java/org/spout/api/protocol/CommonDecoder.java
@@ -79,7 +79,7 @@ public class CommonDecoder extends PreprocessReplayingDecoder {
 		if (codec == null) {
 			return buf;
 		}
-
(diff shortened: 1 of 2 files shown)


zml2008:
Just uncomment https://github.com/SpoutDev/Spout/blob/master/api/src/main/java/org/spout/api/protocol/CommonHandler.java#L84 to print out packets sent/received.

Sleaker:
@zml2008 - does that print opcodes?

zml2008:
nope (opcodes aren't attached to Messages, they're attached to MessageCodecs), but why do you even need opcodes? (but anyway, you shouldn't be PR'ing active debug code this spammy (really, any at all), especially just a plain S.o.println of the codec's opcode).

Sleaker:
Need to be able to debug incoming/outgoing Opcodes on the server to verify that they are matching up properly with the Vanilla client and the official MC server.

PR #2237, opened 2013-08-03
Deleted duplicate dependencies

thehutch:
- Removed unnecessary test console spamming

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -21,12 +21,6 @@
 	<dependencies>
 		<!-- Include in final JAR -->
 		<dependency>
-			<groupId>commons-io</groupId>
-			<artifactId>commons-io</artifactId>
-			<version>2.4</version>
-			<scope>compile</scope>
-		</dependency>
-		<dependency>
 			<groupId>com.google.guava</groupId>
 			<artifactId>guava</artifactId>
 			<version>14.0.1</version>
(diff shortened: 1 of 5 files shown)


lukespragg:
They aren't all duplicate. Each project is using them in different ways, which is perfectly fine. The logging module does not need them all shaded into it, and you didn't actually make it do so with your changes. Either way they aren't hurting anything by being where they are now. Feel free to redo your PR with only the other PR purpose and don't worry about the dependencies.

kitskub:
I don't see any changes...

lukespragg:
I see no changes either. Please fix and correct your commit and PR description. As I said in my previous comment, don't worry about the dependencies, so that means remove the commit for those changes, and only keep the "Removed unnecessary test console spam".

thehutch:
I did make changes, clearly you didn't look at the previous PR much :D. If they're duplicate depends why can't I remove them? I know last time I changed the logging depends to compile instead of provided but I changed that now it only removes the ones which for example are in Cerealization where they are made compile in their pom.xml.

lukespragg:
They aren't duplicated dependencies, they are dependencies used by the associated modules in different ways. Nothing wrong with it, and your changes aren't really what I want to do with them. So for the third time, please do not worry about the dependencies.

I'm just going to close the PR though, again. Feel free to open a new one with just the other change as requested. In the future, please keep unrelated changes in separate PRs so this isn't a problem.

PR #2239, opened 2013-08-04
Improved CodecLookupService

thehutch:
- Changed nextId to an AtomicInteger
- General improvements
- Minor changes to various classes

### DIFF
--- a/api/src/main/java/org/spout/api/protocol/CodecLookupService.java
+++ b/api/src/main/java/org/spout/api/protocol/CodecLookupService.java
@@ -30,8 +30,9 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.lang3.tuple.Pair;
 import org.spout.api.Spout;
@@ -43,22 +44,42 @@ import org.spout.api.util.SyncedStringMap;
  * A class used to lookup message codecs.
  */
 public class CodecLookupService implements EventableListener<SyncedMapEvent> {
-	protected final MessageCodec<?>[] opcodeTable;
-	private final ClassLoader loader;
-	private final SyncedStringMap dynamicPacketMap;
-
-	protected CodecLookupService(ClassLoader loader, SyncedStringMap dynamicPacketMap, int size) {
-		this.dynamicPacketMap = dynamicPacketMap;
-		this.loader = loader;
-		dynamicPacketMap.registerListener(this);
-		opcodeTable = new MessageCodec<?>[size];
-	}
-
-	private int nextId = 0;
 	/**
-	 * A table which maps messages to codecs. This is generally used to map outgoing packets to a codec.
+	 * A lookup table for the Message classes mapped to their Codec.
 	 */
-	protected final Map<Class<? extends Message>, MessageCodec<?>> classTable = new ConcurrentHashMap<>();
+	private final ConcurrentMap<Class<? extends Message>, MessageCodec<?>> classTable;
+	/**
+	 * A synced map for the dynamic packets.
+	 */
+	private final SyncedStringMap dynamicPacketMap;
+	/**
+	 * Lookup table for opcodes mapped to their codecs.
+	 */
+	private final MessageCodec<?>[] opcodeTable;
+	/**
+	 * Stores the next opcode available.
+	 */
+	private final AtomicInteger nextId;
+	/**
+	 * The {@link ClassLoader} for the codecs.
+	 */
+	private final ClassLoader loader;
+
+	/**
+	 * The {@link CodecLookupService} stores the codecs available in the protocol.
+	 * Codecs can be found using either the class of the message they represent or their message's opcode.
+	 *
+	 * @param loader The class loader for the codecs
+	 * @param dynamicPacketMap - The dynamic opcode map
+	 * @param size The maximum number of message types
+	 */
+	protected CodecLookupService(ClassLoader loader, SyncedStringMap dynamicPacketMap, int size) {
+		this.classTable = new ConcurrentHashMap<>(size, 1.0f);
+		this.opcodeTable = new MessageCodec<?>[size];
+		this.dynamicPacketMap = dynamicPacketMap;
+		this.nextId = new AtomicInteger(0);
+		this.loader = loader;
+	}
 
 	/**
 	 * Binds a codec by adding entries for it to the tables.
@@ -71,6 +92,7 @@ public class CodecLookupService implements EventableListener<SyncedMapEvent> {
 	 * @throws InstantiationException if the codec could not be instantiated.
 	 * @throws IllegalAccessException if the codec could not be instantiated due to an access violation.
 	 */
+	@SuppressWarnings("unchecked")
 	protected <T extends Message, C extends MessageCodec<T>> C bind(Class<C> clazz) throws InstantiationException, IllegalAccessException, InvocationTargetException {
 		boolean dynamicId = false;
 		Constructor<C> constructor;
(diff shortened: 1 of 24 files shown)


api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -43,23 +44,18 @@
  * A class used to lookup message codecs.
  */
 public class CodecLookupService implements EventableListener<SyncedMapEvent> {
-	protected final MessageCodec<?>[] opcodeTable;
+	private final TIntObjectMap<MessageCodec<?>> opcodeTable;
+;	private final SyncedStringMap dynamicPacketMap;
```
kitskub: Umm

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -135,11 +117,16 @@ private int getNextId() {
 	 */
 	@SuppressWarnings("unchecked")
 	public <T extends Message> MessageCodec<T> find(Class<T> clazz) {
-		return (MessageCodec<T>) classTable.get(clazz);
+		for(MessageCodec<?> codec : getCodecs()) {
```
kitskub: To be honest, I don't like this. You're trading a map lookup and the small amount of memory from storing that for iterating on every find.
If this method was hardly ever called, this might be a reasonable trade-off (I say might, but probably not). But, this method is call a LOT, so the increase in call time scales. I vote no for this part.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -26,12 +26,13 @@
  */
 package org.spout.api.protocol;
 
+import gnu.trove.map.TIntObjectMap;
+import gnu.trove.map.hash.TIntObjectHashMap;
 import java.lang.reflect.Constructor;
```
kitskub, line 29: Blank line missing

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -26,12 +26,13 @@
...
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.lang3.tuple.Pair;
 import org.spout.api.Spout;
```
kitskub, line 38: Add blank line here please

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -91,39 +90,26 @@ protected CodecLookupService(ClassLoader loader, SyncedStringMap dynamicPacketMa
...
-		codec = register(codec);
-
+		register(codec);
 		dynamicPacketMap.register(clazz.getName(), codec.getOpcode());
 		return codec;
 	}
 
-	private <T extends Message, C extends MessageCodec<T>> C register(C codec) {
+	private <T extends Message, C extends MessageCodec<T>> void register(C codec) {
 		opcodeTable[codec.getOpcode()] = codec;
```
kitskub, line 143: Just keep it returning `C`

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -156,14 +142,16 @@ public void onEvent(SyncedMapEvent event) {
 						throw new IllegalStateException("Server sent a packet id which is greater than the client has allowed.");
 					}
 					 // If the packet has already been registered non-dynamically, don't override
-					if ((opcodeTable[id] != null && opcodeTable[id].getClass().getName().equals(clazzName))) {
+
```
kitskub: Lose this line

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
thehutch, line 143: There's no need for that at all.

kitskub, line 143: Just do it. What's the harm? A plugin could do `registerMine(CodecLookupService.register(Codec));` rather than `Codec; CodecLookupService.register(Codec); registerMine(Codec);`

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -91,39 +114,40 @@ protected CodecLookupService(ClassLoader loader, SyncedStringMap dynamicPacketMa
...
 	}
 
+	/**
+	 * Retrieves the {@link MessageCodec} from the lookup table
+	 * 
+	 * @param opcode The opcode which the codec uses
+	 * @return The codec, null if not found.
+	 */
 	public MessageCodec<?> find(int opcode) {
-		if (opcode > -1 && opcode < opcodeTable.length) {
```
zml2008, line 123: Why are you removing this check? This can cause issues with how we read packet id's.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -155,15 +189,11 @@ public void onEvent(SyncedMapEvent event) {
 					if (id >= opcodeTable.length) {
 						throw new IllegalStateException("Server sent a packet id which is greater than the client has allowed.");
 					}
-					 // If the packet has already been registered non-dynamically, don't override
-					if ((opcodeTable[id] != null && opcodeTable[id].getClass().getName().equals(clazzName))) {
```
zml2008, line 159: This prevents reregistration of packets that are already in the right place and allows non-dynamic packets to work. Don't remove this.

kitskub, line 159: mmm, I partially disagree. The check should stay but the continue should be an error. The server knows about the static opcodes as well. If we receive a dynamic opcode where a static should be, throw an error.

zml2008, line 159: Well then we're going to have to fix the StringMap to correctly store with the id's we tell it to and only register dynamic packets with the map. Currently it seems to override what we give it with autonumbering.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
thehutch, line 123: If you're trying to find illegal packet numbers the developer should know about this, not hide the problem.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
kitskub, line 159: You are correct, zml. The bind method really needs a makeover. I do have two TODOs up there, but there are other things that need fixing. For example, the `codec.setOpcode(nextId.getAndIncrement());` is completely wrong for static opcodes.

EDIT: actually, this looks right.
EDIT: actually, I'm right: needs to throw an error

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -43,34 +44,56 @@
...
+	 * The {@link ClassLoader} for the codecs.
+	 */
+	private final ClassLoader loader;
 
+	/**
+	 * The {@link CodecLookupService} stores the codecs available in the protocol.
+	 * Codecs can be found using either the class of the message they represent or their message's
+	 * opcode.
+	 *
+	 * @param loader           The class loader for the codecs
```
kitskub: We don't format javadocs like this.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
kitskub, line 123: Add the inverse of this check and throw an error instead.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
kitskub, line 159: See EDIT above. Do it.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -138,37 +167,44 @@ private int getNextId() {
...
 					try {
-						// new type(id);
-						Class<? extends MessageCodec> clazz = Class.forName(clazzName, true, loader).asSubclass(MessageCodec.class);
-						Constructor<? extends MessageCodec> constr = clazz.getConstructor(int.class);
-						MessageCodec<?> codec = constr.newInstance(id);
+						final Class<? extends MessageCodec> clazz = Class.forName(className, true, loader).asSubclass(MessageCodec.class);
+						final Constructor<? extends MessageCodec> constr = clazz.getConstructor(int.class);
+						final MessageCodec<?> codec = constr.newInstance(id);
 
 						codec.setDynamic(true);
```
kitskub: Remove this.

api/src/main/java/org/spout/api/protocol/CommonPipelineFactory.java
```diff
@@ -72,7 +67,7 @@ public ChannelPipeline getPipeline() throws Exception {
 		// Up for encoding/sending/downstream; Down for decoding/receiving/upstream
 		CommonEncoder encoder = new CommonEncoder(onClient);
 		CommonDecoder decoder = new CommonDecoder(onClient);
-		CommonHandler handler = new CommonHandler(engine, encoder, decoder);
+		CommonHandler handler = new CommonHandler( encoder, decoder);
```
kitskub: Extra space.

api/src/main/java/org/spout/api/protocol/MessageCodec.java
```diff
@@ -32,8 +32,8 @@
 
 public abstract class MessageCodec<T extends Message> {
 	private final Class<T> clazz;
-	private int opcode;
 	private boolean dynamic;
+	private int opcode;
```
kitskub, line 36: Unnecessary diff.

engine/src/main/java/org/spout/engine/component/entity/ClientTextModelComponent.java
```diff
@@ -82,20 +82,20 @@ public void updateMesh() {
...
 
 		mesh = new BaseMesh(faces, false, true, true, false);
 		mesh.batch();
 	}
 
 	public void render(Camera camera) {
-		if (dirty) {
-			dirty = false;
+		if (isDirty()) {
+			setDirty(true);
```
kitskub: Bad logic here.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
```diff
@@ -84,46 +106,56 @@ protected CodecLookupService(ClassLoader loader, SyncedStringMap dynamicPacketMa
...
+	 * Retrieves the {@link MessageCodec} from the lookup table
+	 *
+	 * @param opcode The opcode which the codec uses
+	 * @return The codec, null if not found.
+	 */
 	public MessageCodec<?> find(int opcode) {
-		if (opcode > -1 && opcode < opcodeTable.length) {
-			return opcodeTable[opcode];
+		//TODO: Remove pointless bounds check
+		if (opcode < 0 || opcode > opcodeTable.length + 1) {
```
kitskub: why not `if (opcode < 0 || opcode >= opcodeTable.length) {`

And remove the TODO

thehutch: I'll fix the pointless bounds check, the TODO can stay since it's a change that will need to occur

api/src/main/java/org/spout/api/protocol/MessageCodec.java
thehutch, line 36: erm no

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
kitskub: No, it's not a change that will need to occur. Or will occur. Remove it.

api/src/main/java/org/spout/api/protocol/CodecLookupService.java
macalinao, line 38: :-1: James tbh most of your pulls don't add anything, they're all minor
changes that change code to your own style

kitskub, line 38: Ian tbh most of your comments don't add anything, they're all mean
remarks that are off topic.

But seriously: while a lot of his changes are style/formatting changes, he has submitted a lot of good code. And formatting is NOT bad. If the only reason you're commenting is to give NEGATIVE comments (not even feedback, just comments) then go away.

macalinao, line 38: Ok here's my feedback. You are making microoptimizations on code that is
not released yet. It is pointless and this code is probably going to be
changed very frequently until a beta.

kitskub, line 38: There. Was that so hard?

PR #2241, opened 2013-08-06
Restructured SpoutRegion from ground up

thehutch:

### DIFF
--- a/api/src/main/java/org/spout/api/util/list/concurrent/setqueue/SetQueueElement.java
+++ b/api/src/main/java/org/spout/api/util/list/concurrent/setqueue/SetQueueElement.java
@@ -31,7 +31,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 /**
  * An element for a SetQueue.<br> <br> Calls to the add() method have no effect if the element is already in the queue.
  */
-public abstract class SetQueueElement<T> {
+public class SetQueueElement<T> {
 	private final SetQueue<T> queue;
 	private final T value;
 	private final AtomicBoolean queued = new AtomicBoolean(false);
@@ -69,7 +69,9 @@ public abstract class SetQueueElement<T> {
 		return false;
 	}
 
-	protected abstract boolean isValid();
+	protected boolean isValid() {
+		return true;
+	}
 
 	@Override
 	public String toString() {
--- a/engine/src/main/java/org/spout/engine/world/SpoutChunk.java
+++ b/engine/src/main/java/org/spout/engine/world/SpoutChunk.java
@@ -271,14 +271,14 @@ public class SpoutChunk extends Chunk implements Snapshotable, Modifiable {
 		// loaded chunk
 		selfReference = new WeakReference<>(this);
 		this.scheduler = (SpoutScheduler) Spout.getScheduler();
-		this.saveMarkedElement = new ChunkSetQueueElement<>(getRegion().saveMarkedQueue, this);
-		this.unloadQueueElement = new ChunkSetQueueElement<>(getRegion().unloadQueue, this);
-		this.populationQueueElement = new ChunkSetQueueElement<>(getRegion().populationQueue, this);
-		this.populationPriorityQueueElement = new ChunkSetQueueElement<>(getRegion().populationPriorityQueue, this);
-		this.localPhysicsChunkQueueElement = new ChunkSetQueueElement<>(getRegion().localPhysicsChunkQueue, this);
-		this.globalPhysicsChunkQueueElement = new ChunkSetQueueElement<>(getRegion().globalPhysicsChunkQueue, this);
-		this.dirtyChunkQueueElement = new ChunkSetQueueElement<>(getRegion().dirtyChunkQueue, this);
-		this.newChunkQueueElement = new ChunkSetQueueElement<>(getRegion().newChunkQueue, this);
+		this.saveMarkedElement = new ChunkSetQueueElement<>(region.markedSaveQueue, this);
+		this.unloadQueueElement = new ChunkSetQueueElement<>(region.unloadQueue, this);
+		this.populationQueueElement = new ChunkSetQueueElement<>(region.populationQueue, this);
+		this.populationPriorityQueueElement = new ChunkSetQueueElement<>(region.populationPriorityQueue, this);
+		this.localPhysicsChunkQueueElement = new ChunkSetQueueElement<>(region.localPhysicsChunkQueue, this);
+		this.globalPhysicsChunkQueueElement = new ChunkSetQueueElement<>(region.globalPhysicsChunkQueue, this);
+		this.dirtyChunkQueueElement = new ChunkSetQueueElement<>(region.dirtyChunkQueue, this);
+		this.newChunkQueueElement = new ChunkSetQueueElement<>(region.newChunkQueue, this);
 	}
 
 	@Override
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/spout/api/util/list/concurrent/setqueue/SetQueueElement.java
```diff
@@ -31,10 +31,10 @@
 /**
  * An element for a SetQueue.<br> <br> Calls to the add() method have no effect if the element is already in the queue.
  */
-public abstract class SetQueueElement<T> {
+public class SetQueueElement<T> {
```
kitskub, line 34: Why unabstract it?

engine/src/main/java/org/spout/engine/world/SpoutChunk.java
```diff
@@ -298,30 +298,30 @@ private SpoutChunk(SpoutWorld world, SpoutRegion region, float x, float y, float
...
-		this.saveMarkedElement = new ChunkSetQueueElement<>(getRegion().saveMarkedQueue, this);
-		this.unloadQueueElement = new ChunkSetQueueElement<>(getRegion().unloadQueue, this);
-		this.populationQueueElement = new ChunkSetQueueElement<>(getRegion().populationQueue, this);
-		this.populationPriorityQueueElement = new ChunkSetQueueElement<>(getRegion().populationPriorityQueue, this);
-		this.chunkObserversDirtyQueueElement = new ChunkSetQueueElement<>(getRegion().chunkObserversDirtyQueue, this, true);
-		this.localPhysicsChunkQueueElement = new ChunkSetQueueElement<>(getRegion().localPhysicsChunkQueue, this);
-		this.globalPhysicsChunkQueueElement = new ChunkSetQueueElement<>(getRegion().globalPhysicsChunkQueue, this);
-		this.dirtyChunkQueueElement = new ChunkSetQueueElement<>(getRegion().dirtyChunkQueue, this);
-		this.newChunkQueueElement = new ChunkSetQueueElement<>(getRegion().newChunkQueue, this);
+		this.saveMarkedElement = new ChunkSetQueueElement<>(region.getMarkedSaveQueue(), this);
```
kitskub: I don't see a reason to change these. They're not API public.

engine/src/main/java/org/spout/engine/world/SpoutChunk.java
```diff
@@ -863,8 +863,8 @@ public boolean removeObserver(Entity entity) {
 		chunkObserversDirtyQueueElement.add();
 		TickStage.checkStage(TickStage.FINALIZE);
 
-		if (observers.remove(entity) && (entity instanceof SpoutPlayer)) {
-			observingPlayers.remove(entity);
+		if (entity instanceof SpoutPlayer && observers.remove((SpoutPlayer) entity)) {
```
kitskub: This logic is different. `observers` contains entities other than players. The new code doesn't account for that.

api/src/main/java/org/spout/api/util/list/concurrent/setqueue/SetQueueElement.java
thehutch, line 34: I wanted to remove RegionSetQueueElement or something like that and so this needed to become non abstract

engine/src/main/java/org/spout/engine/world/SpoutChunk.java
thehutch: You shouldn't really access fields directly from another class and also getRegion() is overrideable and Netbeans was telling me to change it, also getRegion() uses field calls so using the local variable is quicker

engine/src/main/java/org/spout/engine/world/SpoutChunk.java
thehutch: Observers contains SpoutPlayers not entities.
`LinkedHashMap<SpoutPlayer, TByteTripleHashSet> observers = new LinkedHashMap<>()`

kitskub: uh, no it doesn't `private final Set<SpoutEntity> observers = Sets.newSetFromMap(new ConcurrentHashMap<SpoutEntity, Boolean>());`

engine/src/main/java/org/spout/engine/world/SpoutChunk.java
kitskub: then make getRegion() final. Honestly, too much boilerplate code.

engine/src/main/java/org/spout/engine/world/SpoutChunk.java
thehutch: oops, was looking in SpoutRegion :D

engine/src/main/java/org/spout/engine/world/SpoutChunk.java
thehutch: (facepalm)

kitskub:
This PR adds 300+ lines of boilerplate code to SpoutRegion and removes one-time-use methods for the sake of "performance" versus readability. Multiple `for` statement spacing is off in SpoutRegion.

I personally vote no for this.

thehutch:
I vote that @kitskub should have no vote in this PR. No offence but he is not being nice to me or my work. If there are sensible reasons why I need to make changes then I'll make them, don't just say "boilerplate" and expect that to be a reasonable argument.

kitskub:
Let me be VERY clear, @thehutch. When I opened your PR to review it for the first time, I had NO intention to purposely give you a hard time over your code. However, if I don't like the code, I'm going to say something, not let it sit there. Don't confuse criticism with meanness.
I gave you sensible reasons:
1) boilerplate. SpoutRegion is 340+ lines longer with this PR. There is little to minimal logic changes/additions. This means that those 340+ lines do exactly what the previous code did, except now 340+ lines longer. The protected variables are PROTECTED. This is the engine, not the API; we don't need to worry about plugins modifying the code.
2) readability. You've removed many "one-time-use" methods. While they are functionally unneeded, they allow a visual separation and blocking of code blocks. You say you remove them for "performance reasons." To that, I link this thread: http://stackoverflow.com/questions/1830179/what-is-the-overhead-of-a-method-call-in-a-good-java-vm
3) maintability. With the additional 340+ lines of boilerplate comes an additional 340+ lines of code to maintain (even though getters are simple to maintain). For all those removed methods, logic blocks become less apparent than as separate methods.

As far as the moving of initializing variables to the constructor, I'm not entirely sure if there are any formatting guidelines for me. But if there isn't, then I link this thread to backup my point that they should stay with declaration: http://stackoverflow.com/questions/3918578/should-i-initialize-variable-within-constructor-or-outside-constructor

Again, I'm NOT out to get you or out to be mean. I'm here to improve Spout, including commenting on PRs that contain, in my honest opinion, ludicrous changes.

There is plenty of GOOD in the PR that I don't comment on. My comments aren't to pat you on the back for good code, it's to dispute code that might need changing.

thehutch:
1) Given the fact that I didn't add any extra code, I believe that the so called "boilerplate" code comes mostly from the Javadocs I added. Yes I think that get methods are better, they make it more maintainable, not the other way around.

2) So I've removed lines of code and somehow it's bigger :P

3) I'm pretty sure that thread agrees more with putting the initialisation in the constructor, not at the field level.

kitskub:
@thehutch I'm only going to resond to the third one because that's the only concrete one to rebute.
No, multiple people say declaration and initialization together is better:
"I find the second style (declaration + initialization in one go) superior."
"I tend to use the second one to avoid a complicated constructor (or a useless one),"
But there are people that like the other way.
Again, it's preference. And I personally side with declaration + initialization at the same time.

kitskub:
@thehutch this needs a rebase :)

thehutch:
@kitskub I have rebased this now :D

kitskub:
@zidane will have to review this one because can't view it web-only and i don't have a computer.

kitskub:
I'm so sorry @thehutch, but I'm going to close this PR. If you would like to resubmit, you can open another Pull Request. And if you do, it would be AMAZING if you can split up the changes so that we can review the changes from the web. Thanks anyways!

thehutch:
Well I have just started university now, but if I find some time, I can do it. I do have a copy of the work anyway.

PR #2242, opened 2013-08-07
Eliminate duplicate pom.xml code by using a named property

kitskub:
@Wulfspider I didn't think you would want me touching the pom.xml without your eyes. So I made a PR.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -17,6 +17,10 @@
 		<version>1</version>
 	</parent>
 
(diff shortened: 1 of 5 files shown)


macalinao:
this is cool :+1:

lukespragg:
Smart idea @kitskub. I have other changes to push, and got a couple things I'll change with this so will set it up locally. :)

PR #2247, opened 2013-08-18
Only load .jar . We don't want anything else (.DS_STORE on macs, etc.)

greatman:
Signed-off-by: William Lebel lebelwilliam@gmail.com


Wolf480pl:
Shouldn't it be endsWith() ?

windy1:
Yeah, what if I have a `file.jar.zip`? It would still be accepted.

DDoS:
@W1ndwaker Spout kept trying to load system files as plugins (DS_STORE files).

PR #2249, opened 2013-08-29
Remove unnecessary 'abstract' modifiers (Widget.java)

olliestanley:


Zidane:
GUI API is going into a new branch and this will change

PR #2250, opened 2013-09-04
Changed mesh thread from Set to ArrayBlockingQueue

Waterpicker:


DDoS:
_facepalm_

@kitskub you should've been more specific.

@Waterpicker we were talking about the chunk queue for meshing, not the thread collection.

kitskub:
Hahaha. I definitely should have been more specific. Thanks for trying anyways. I'm free again, so I'll just do it.

PR #2251, opened 2013-09-17
Adds hasJoinedBefore and PlayerFirstJoinEvent. SPOUT-220

Amperial:


lukespragg:
In addition to the formatting comments I made, you'd need to sign-off on your commit.

Sign-off, verifying agreement with the license terms.

A commit with no detailed description can be created using:
git commit -s -m "A brief description of the change"

Amperial:
Fixed.

zml2008:
Having a separate event is stupid. If statements exist for a reason.

Amperial:
It seems like @kitskub @thehutch and Ian think it would be better to have a separate one, I kinda agree.
I can make it one event again if you guys want though?

zml2008:
The subclass adds no methods, only exists for `event instance of PlayerFirstJoin`, which can be done just as well with `!player.hasJoinedBefore()`

Amperial:
What if I added a get/set welcome message to the first join event, that gets sent right before the join message?
That way it would actually serve a purpose, and servers wouldn't need a plugin to take care of that.

Grinch:
Vanilla isn't set out to replace plugins, just to provide an extensible API. This event (in my opinion) isn't really needed as they can simply do a `if (!hasJoinedBefore)` check during the PlayerJoinEvent. :/

kitskub:
As long hasJoinedBefore is false during the event, then no need for new event.

engine/src/main/java/org/spout/engine/entity/SpoutPlayer.java
```diff
@@ -89,8 +90,12 @@ public SpoutPlayer(Engine engine, Class<? extends PlayerNetworkComponent> networ
 		copySnapshot();
 	}
 
+	/**
+	 * Only used when creating a new SpoutPlayer, hence firstLogin being set to true.
+	 */
 	public SpoutPlayer(Engine engine, Class<? extends PlayerNetworkComponent> network, String name, Transform transform) {
 		this(engine, network, name, transform, null, (byte[]) null, (Class<? extends Component>[]) null);
+	    firstLogin = true;
```
kitskub: Formatting...

api/src/main/java/org/spout/api/entity/Player.java
```diff
@@ -68,6 +68,13 @@
 	public boolean isOnline();
 
 	/**
+	 * Gets if the player has joined before
+	 *
```
kitskub: Since you have to edit the commit for formatting below, can you remove this space? Thanks!

lukespragg: Trailing whitespace, not the new line. ^

engine/src/main/java/org/spout/engine/entity/SpoutPlayer.java
lukespragg: By formatting, he means we use tabs, not spaces.

Amperial:
Yeah i'll fix it tonight, sorry :P
Usually I just auto format and it removes the spaces, but that would change a bunch of other stuff so I had to do it manually.

engine/src/main/java/org/spout/engine/entity/SpoutPlayer.java
kitskub: Si

api/src/main/java/org/spout/api/entity/Player.java
kitskub: :dart:

Amperial:
Fixed the two formatting errors.

kitskub:
One last small thing (sorry!), can you edit the commit description to not include the " and PlayerFirstJoinEvent"?

Thanks! Really! That's the only thing I see!

Amperial:
Alright, done :) @kitskub

kitskub:
thanks!

PR #2252, opened 2013-09-17
created TextureLoader and RenderMaterialLoader

Waterpicker:

### DIFF
--- a/engine/src/main/java/org/spout/engine/filesystem/resource/loader/RenderMaterialLoader.java
+++ b/engine/src/main/java/org/spout/engine/filesystem/resource/loader/RenderMaterialLoader.java
@@ -1,0 +1,192 @@
+/*
+ * To change this template, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.spout.engine.filesystem.resource.loader;
+
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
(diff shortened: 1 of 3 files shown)


Waterpicker:
If you think it won't work as is let me know, this is just so we something to work with.

engine/src/main/java/org/spout/engine/filesystem/resource/loader/RenderMaterialLoader.java
```diff
@@ -0,0 +1,192 @@
+/*
```
kitskub, line 1: ...

engine/src/main/java/org/spout/engine/filesystem/resource/loader/RenderMaterialLoader.java
```diff
@@ -0,0 +1,192 @@
...
+import org.spout.math.vector.Vector4;
+import org.spout.renderer.Material;
+import org.spout.renderer.data.Color;
+import org.spout.renderer.gl.Program;
+import org.spout.renderer.gl.Texture;
+import org.yaml.snakeyaml.Yaml;
+
+/**
+ *
+ * @author Johnson
```
kitskub, line 53: ...

DDoS:
This won't work. You can't create the GL objects during loading (the GL context won't always be created). The loader shouldn't even be loading GL objects from Caustic (and yes, I know the current shader loader does this, I have to fix that), we need wrappers for the API anyways.

PR #2253, opened 2013-09-18
Texture updated

Waterpicker:


api/src/main/java/org/spout/api/render/Texture.java
```diff
@@ -26,75 +26,27 @@
...
+		return edge.getFloorX();
 	}
 
-	public final int getWidth() {
-		return this.width;
+	public int getHeight() {
+		return edge.getFloorY();
 	}
-
+
```
kitskub: there should not be a tab here

DDoS:
1. Doesn't compile
2. Needs to be squashed
3. Image data should only be stored in the caustic texture. Right now it's not even being disposed of after uploading to the GPU which will leave a potentially large amount of useless data in the memory.

kitskub:
@DDoS 
1) I wouldn't judge the "doesn't compile" based solely of Travis because of how many problems it gives up.
2) I'm being pretty lenient with the squashing because I don't mind pulling it locally and squashing myself if it means we get some code.

Waterpicker:
@DDoS 
1. Agreed with kitskub but I do need to test.
2. Thats due to my confusion about how caustic objects are made.

PR #2254, opened 2013-09-19
Texture Loading complete

Waterpicker:
Waterpicker Watepricker@rocketmail.com

### DIFF
--- a/engine/src/main/java/org/spout/engine/filesystem/resource/SpoutTexture.java
+++ b/engine/src/main/java/org/spout/engine/filesystem/resource/SpoutTexture.java
@@ -1,0 +1,147 @@
+/*
+ * This file is part of Spout.
+ *
+ * Copyright (c) 2011 Spout LLC <http://www.spout.org/>
+ * Spout is licensed under the Spout License Version 1.
+ *
+ * Spout is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 3 of the License, or (at your option)
+ * any later version.
+ *
+ * In addition, 180 days after any changes are published, you can use the
+ * software, incorporating those changes, under the terms of the MIT license,
+ * as described in the Spout License Version 1.
+ *
+ * Spout is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License,
+ * the MIT license and the Spout License Version 1 along with this program.
+ * If not, see <http://www.gnu.org/licenses/> for the GNU Lesser General Public
+ * License and see <http://spout.in/licensev1> for the full license, including
+ * the MIT license.
+ */
+package org.spout.engine.filesystem.resource;
+
+import java.awt.image.BufferedImage;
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/spout/engine/filesystem/resource/SpoutTexture.java
```diff
@@ -0,0 +1,141 @@
...
+import org.spout.renderer.gl.Texture.Format;
+import org.spout.renderer.gl.Texture.InternalFormat;
+import org.spout.renderer.gl.Texture.WrapMode;
+import org.spout.renderer.util.CausticUtil;
+import org.spout.renderer.util.Rectangle;
+
+public class SpoutTexture extends Texture {
+	org.spout.renderer.gl.Texture texture;
+	
+	public SpoutTexture(BufferedImage baseImage, Vector2 edge) {
```
DDoS: The edge here is redundant, it's stored in the buffered image

engine/src/main/java/org/spout/engine/filesystem/resource/SpoutTexture.java
```diff
@@ -0,0 +1,141 @@
...
+	public void setComponentType(DataType type) {
+		texture.setComponentType(type);
+	}
+	
+	public void setFormat(Format format) {
+		texture.setFormat(format);
+	}
+	
+	public void setImageData(ByteBuffer buffer, int x, int y) {
+		texture.setImageData(buffer, x, y);
```
DDoS: This should be width and height, not x and y, to avoid confusion.

DDoS:
This still keeps the image data in the memory. You store a buffered image in the texture class, and set the data to the caustic object just after. The caustic object will cleanup the data after creation, but not the texture class. You shouldn't be setting the image data in the caustic object before creation, and after you should set the buffered image to null to let it be GC'ed. Once the data has been uploaded to the GPU after creation, we shouldn't need to have on the memory.

Also, formatting is wrong. You're using Netbeans right? You know there's a key command to format your file automatically... If the formatting Netbeans is using is wrong, you can change it in the preferences.

kitskub:
@DDoS what's wrong with formatting?

engine/src/main/java/org/spout/engine/filesystem/resource/SpoutTexture.java
```diff
@@ -0,0 +1,140 @@
...
+	}
+	
+	public void setWrapS(WrapMode mode) {
+		texture.setWrapS(mode);
+	}
+	
+	public void setWrapT(WrapMode mode) {
+		texture.setWrapT(mode);
+	}
+	public void unbind() {
```
DDoS, line 124: This

engine/src/main/java/org/spout/engine/filesystem/resource/SpoutTexture.java
```diff
@@ -0,0 +1,140 @@
...
+	}
+	
+	public void checkCreated() {
+		texture.checkCreated();
+	}
+	
+	public void create() {
+		texture.create();
+		
+	}
```
DDoS, line 67: This

lukespragg, line 67: New line not needed at line 69.

engine/src/main/java/org/spout/engine/filesystem/resource/SpoutTexture.java
lukespragg, line 124: New line needed after line 125, before 126.

PR #2255, opened 2013-09-21
Shader and Program update

Waterpicker:
Signed-off-by: Waterpicker waterpicker@rocketmail.com

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -38,6 +38,12 @@
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.fourthline.cling</groupId>
+			<artifactId>cling-support</artifactId>
+			<version>2.0-alpha2</version>
+			<scope>compile</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.spout</groupId>
 			<artifactId>caustic</artifactId>
 			<version>1.0.0-SNAPSHOT</version>


lukespragg:
Also, not sure if you are manually inserting your sign-off line or not, but it should be:

Signed-off-by: Waterpicker <waterpicker@rocketmail.com>

Waterpicker:
Sorta fixed it.

lukespragg:
Having the sign-off in the PR comments is not the same as the commit itself by the way. ;)

api/src/main/java/org/spout/api/render/shader/Program.java
```diff
@@ -0,0 +1,59 @@
...
+
+	public abstract void setUniform(String string, Vector2[] vector);
+
+	public abstract void setUniform(String string, Vector3 vector);
+
+	public abstract void setUniform(String string, Vector3[] vector);
+
+	public abstract void setUniform(String string, Vector4 vector);
+
+	public abstract void setUniform(String string, Boolean bool);
```
DDoS: Why is this a boxed type? (Boolean instead of boolean)

engine/src/main/java/org/spout/engine/filesystem/resource/SpoutShader.java
```diff
@@ -0,0 +1,82 @@
...
+	public boolean isCreated() {
+		return shader.isCreated();
+	}
+
+	public void setSource(InputStream in) {
+		shader.setSource(in);
+	}
+
+	public void setType(ShaderType type){
+		shader.setType(Shader.ShaderType.VERTEX);
```
DDoS: This doesn't work...

engine/src/main/java/org/spout/engine/filesystem/resource/loader/ProgramLoader.java
```diff
@@ -70,25 +70,26 @@ public Program load(InputStream in) {
 			throw new IllegalStateException("Missing version in spout shader file: " + mode);
 		}
 
-		final Shader vertex = client.getRenderer().getGL().createShader();
+		final SpoutShader vertex;
 		String shaderFile = shaderNode.getNode("Vertex").getString();
 		InputStream shaderSource = Spout.getFileSystem().getResourceStream(shaderFile);
```
DDoS: This should be modified to load shaders directly with the new shader loader, instead of loading a source and constructing the shader manually. Right now, the way this is setup, the shader loader is useless.

engine/src/main/java/org/spout/engine/filesystem/resource/loader/ShaderLoader.java
```diff
@@ -0,0 +1,46 @@
...
+	public ShaderLoader() {
+		super("shader", null);
+	}
+
+	@Override
+	public SpoutShader load(InputStream in) {
+		return new SpoutShader(in);
+	}
+
+}
```
DDoS: Formatting, there should be no new line at 45.

DDoS:
I can see right now that figuring out the shader type (vertex, fragment, etc...) is a problem. Since we only have the input stream, we can't check the file extension. We could use multiple shader loaders, one for each type, but that doesn't seem efficient at all. I think I'll add a token to the Caustic shader parser, which will enable the developers to specify the shader type in the source. Caustic will set the shader type automatically if the token is present.

This means that some extra information will be required in the shaders for Spout, but I don't think it's going to be a huge problem. I have already made the current texture and attribute tokens mandatory, to simplify shader loading.

PR #2259, opened 2013-10-03
removed Camera component and added Camera Wrapper

Waterpicker:


engine/src/main/java/org/spout/engine/renderer/SpoutCamera.java
```diff
@@ -0,0 +1,86 @@
...
+ * You should have received a copy of the GNU Lesser General Public License,
+ * the MIT license and the Spout License Version 1 along with this program.
+ * If not, see <http://www.gnu.org/licenses/> for the GNU Lesser General Public
+ * License and see <http://spout.in/licensev1> for the full license, including
+ * the MIT license.
+ */
+/*
+ * To change this template, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
greatman: Forgot the auto-commenter ;)

engine/src/main/java/org/spout/engine/SpoutClient.java
```diff
@@ -447,4 +455,14 @@ public SpoutScreenStack getScreenStack() {
 	public SpoutRenderer getRenderer() {
 		return renderer;
 	}
+
+	@Override
+	public void setCamera(Camera camera) {
+		this.camera = (SpoutCamera) camera;
```
DDoS, line 461: This cast will prevent any camera implementation other than Spout's integrated one.

engine/src/main/java/org/spout/engine/renderer/SpoutCamera.java
```diff
@@ -0,0 +1,83 @@
...
+import org.spout.math.imaginary.Quaternion;
+import org.spout.math.matrix.Matrix4;
+import org.spout.math.vector.Vector3;
+
+public class SpoutCamera extends org.spout.renderer.Camera implements Camera {
+
+	public SpoutCamera(Matrix4 projection) {
+		super(projection);
+	}
+	@Override
```
DDoS, line 40: Formatting. Also, extending the camera and implementing an interface with the exact same methods feels like a weird design choice to me, I would prefer if this was only a wrapper, it would be less confusing.

engine/src/main/java/org/spout/engine/SpoutRenderer.java
```diff
@@ -218,22 +218,4 @@ private void generateChunkModels() {
 
 		chunkMeshes.clear();
 	}
-
```
DDoS, line 221: You broke the rendering because you broke input. You need to have the camera in the Client reflect the player's transform (and have it updated on each frame). Also, the camera should be in the renderer, not the client.

PR #2263, opened 2013-12-25
make mapHash return something other than 0

mebigfatguy:
as written

 int intKey = (key >> 32 ^ key);

creates 0, as shifting by 32 is the same as shifting by 0, then exclusive oring with self is 0. always 0.

Make the hash distributed, by adding parens

int intKey = (key >> (32 ^ key));


kitskub:
@mebigfatguy What kind of implications does this have?

mebigfatguy:
The way it was before, all you map entries where hashed to the same bucket, turning your map into a collection that had List-like performance for lookup, etc. You probably could do even better with that hash function, but certainly it was not good at all as it was, and should be much better.

kitskub:
So you're saying this provides a speed increase?

mebigfatguy:
should, how measurable, hard to say.

